PR #3, opened 2018-07-31
Proposal 0001 - Extract WebView From Core

jamonholmgren:
This proposes that we extract the current WebView implementation from React Native Core and move it into a third party package.

### DIFF
--- a/proposals/0001-template.md
+++ b/proposals/0001-template.md
@@ -25,9 +25,11 @@ This would move it to an external package, like so:
 import { WebView } from '@infinitered/react-native-webview';
 ```
 
+(Where this third-party package will live is an open question. See below.)
+
 ## Motivation
 
-The current WebView implementation is (to my knowledge) not used by Facebook internally. This leads (understandably) to a lack of first-class support for this core API. By moving the component to a third party that is more invested in supporting it, there will be better support.
+The current WebView implementation is used by Facebook internally but not actively updated. This leads (understandably) to a lack of first-class support for this core API. By moving the component to a third party that is more invested in supporting it, there will be better support.
 
 Additionally, with this component no longer included in core, there will be an opening for alternative third party WebView implementations, should anyone else wish to provide one.
 
@@ -57,4 +59,6 @@ We will need to update the React Native documentation to point out that WebView 
 
 ## Unresolved questions
 
-None so far.
+There are a number of [WebView-related issues](https://github.com/facebook/react-native/search?q=webview&state=open&type=Issues) in the core React Native repo. We will need to have a plan to migrate those over to the new WebView repo.
+
+Additionally, we need to decide where this new repo will live. I am willing to host it under the `infinitered` organization (this motivates me more to keep it up to date and supported), but am open to other ideas such as `react-native-community`.
--- a/proposals/0001-template.md
+++ b/proposals/0001-template.md
@@ -1,0 +1,60 @@
+---
+title: Extract WebView From Core
+author:
+- Jamon Holmgren
+date: 2018-07-30
+---
+
+# RFC0000: Extract WebView From Core
+
+## Summary
+
+Extract the current WebView implementation from React Native Core and move it into a third party package.
+
+## Basic example
+
+Currently, WebView is included in React Native core:
+
+```jsx
+import { WebView } from 'react-native';
+```
+
+This would move it to an external package, like so:
+
+```jsx
+import { WebView } from '@infinitered/react-native-webview';
+```
+
+## Motivation
+
+The current WebView implementation is (to my knowledge) not used by Facebook internally. This leads (understandably) to a lack of first-class support for this core API. By moving the component to a third party that is more invested in supporting it, there will be better support.
+
+Additionally, with this component no longer included in core, there will be an opening for alternative third party WebView implementations, should anyone else wish to provide one.
+
+## Detailed design
+
+For the first release, the goal is to provide the same API, only replacing the import statement and otherwise backwards compatible.
+
+We will also want to update generated project boilerplates, such as create-react-native-app and Ignite, to use the new package.
+
+## Drawbacks
+
+This is a breaking change in that if someone simply updates their React Native version in an app that uses WebView and this core API is gone, their app will break.
+
+It also means there won't be an officially supported WebView in React Native, which may feel to some to be missing key functionality.
+
+## Alternatives
+
+The primary alternative is to keep the WebView in core. There are also existing third party WebViews, although none that are reliably cross-platform.
+
+## Adoption strategy
+
+This is a breaking change and existing React Native authors who use WebView will need to update their apps to access one of the third party solutions. For most, it should be as simple as `yarn add @infinitered/react-native-webview` and update their import statements.
+
+## How we teach this
+
+We will need to update the React Native documentation to point out that WebView is now moved to a third party module.
+
+## Unresolved questions
+
+None so far.
--- a/proposals/0001-webview.md
+++ b/proposals/0001-webview.md
@@ -22,7 +22,7 @@ import { WebView } from 'react-native';
 This would move it to an external package, like so:
 
 ```jsx
-import { WebView } from '@infinitered/react-native-webview';
+import { WebView } from 'react-native-webview';
 ```
 
 (Where this third-party package will live is an open question. See below.)
--- a/proposals/0001-webview.md
+++ b/proposals/0001-webview.md
@@ -51,7 +51,7 @@ The primary alternative is to keep the WebView in core. There are also existing 
 
 ## Adoption strategy
 
-This is a breaking change and existing React Native authors who use WebView will need to update their apps to access one of the third party solutions. For most, it should be as simple as `yarn add @infinitered/react-native-webview` and update their import statements.
+This is a breaking change and existing React Native authors who use WebView will need to update their apps to access one of the third party solutions. For most, it should be as simple as `yarn add react-native-webview` and update their import statements.
 
 ## How we teach this
 


proposals/0001-template.md
```diff
@@ -0,0 +1,60 @@
...
+
+This would move it to an external package, like so:
+
+```jsx
+import { WebView } from '@infinitered/react-native-webview';
+```
+
+## Motivation
+
+The current WebView implementation is (to my knowledge) not used by Facebook internally. This leads (understandably) to a lack of first-class support for this core API. By moving the component to a third party that is more invested in supporting it, there will be better support.
```
axe-fb: Actually, we do use WebView internally, but have not been maintaining it as much.

proposals/0001-template.md
```diff
@@ -0,0 +1,60 @@
...
+
+Additionally, with this component no longer included in core, there will be an opening for alternative third party WebView implementations, should anyone else wish to provide one.
+
+## Detailed design
+
+For the first release, the goal is to provide the same API, only replacing the import statement and otherwise backwards compatible.
+
+We will also want to update generated project boilerplates, such as create-react-native-app and Ignite, to use the new package.
```
axe-fb: Should we do it in the first add more features/bigs and then look at a replacement ?

proposals/0001-template.md
```diff
@@ -0,0 +1,60 @@
...
+
+This is a breaking change and existing React Native authors who use WebView will need to update their apps to access one of the third party solutions. For most, it should be as simple as `yarn add @infinitered/react-native-webview` and update their import statements.
+
+## How we teach this
+
+We will need to update the React Native documentation to point out that WebView is now moved to a third party module.
+
+## Unresolved questions
+
+None so far.
```
axe-fb: I think we also need to talk about how we move existing bugs and PRs over to the new location ?

proposals/0001-template.md
```diff
@@ -0,0 +1,60 @@
...
+Currently, WebView is included in React Native core:
+
+```jsx
+import { WebView } from 'react-native';
+```
+
+This would move it to an external package, like so:
+
+```jsx
+import { WebView } from '@infinitered/react-native-webview';
```
axe-fb: What should be the right source ? Should this be in infinitered, or should it be in react-native-community ?

proposals/0001-template.md
jamonholmgren: I'm definitely open to that. I mainly wanted to keep the surface area of the change smaller for those who have to migrate their existing apps.

proposals/0001-template.md
jamonholmgren: Right, I'll add that.

proposals/0001-template.md
jamonholmgren: I'm open to react-native-community, but am more motivated to devote my company resources toward it if it lives in `infinitered`.

kelset:
To give some extra context:
* link to the open issues related to [`WebView`](https://github.com/facebook/react-native/issues?page=1&q=is%3Aissue+is%3Aopen+label%3A%3Alarge_orange_diamond%3AWebView&utf8=%E2%9C%93) in the main repo
* the discussion issue of moving to [WKWebWview instead of UIWebView](https://github.com/facebook/react-native/issues/321)

LinusU:
I'm personally not a fan of moving `WebView` out of core, even though I feel the frustration of submitting PRs that neither gets comments nor gets merged.

My reasoning builds on two things:

**1)** There are a lot of small useful modules which uses WebView behind the scenes for things to work. Some quick examples are [`react-native-qrcode`](https://github.com/cssivision/react-native-qrcode), [`react-native-webview-crypto`](https://github.com/saulshanabrook/react-native-webview-crypto), [`react-native-editor`](https://github.com/nt-team/react-native-editor), etc...

If we move `WebView` out of core, these will no longer "just work" after an `npm install`. Simply declaring a dependency on the `WebView`-module won't work either, since it will require manual setup to work.

Having dependencies on modules that requires manual setup provides a very poor user experience for your users. You basically have to duplicate the other library's setup instructions in your own readme, and hope that the users are paying attention.

Until React Native has a way to import native dependencies which require zero input from the user (kind of like how prebuilt native modules works in Node.js) I don't think it's a good idea to move this out.

and also **2)** there is nothing stopping anyone from developing a user-land `WebView`. Instead of making a hard switch, I think that a better approach is to let one of the user-land implementations prove themselves first, and then we can first deprecate, and then remove, the built-in `WebView`.

Just my 2¢ ☺️

eliperkins:
What's preventing the community from creating a package that does this on it's own first? It feels like we should be providing the implementation first, then deprecating `WebView` and recommending moving to the new package, before we set out down this road.

axe-fb:
I think the goal here is to enable a WebView that is maintained and supported well. I am agnostic to picking any existing community packages, or creating a new one, as long it is well maintained and owned in the community.

I think that we have not been able to give web views enough attention, and thats the reason I would like use to move this out. 

Regarding other modules depending on WebView - I think we don't have a story for React-native modules depending on each other. Folks from Artsy were working on a proposal for better `react-native link`, with cocoapods, but if there are other proposals that the community has, lets add them to this repo.

grabbou:
The `link` itself should work just fine. I am actually working on a new proposal (as a part of this repository) to kick start discussion on the new format and dependency management.

Right now, if the module depends solely on its own files and contains no extraneous native dependencies, things work just fine. We also support use of Cocoapods. On Android this is not an issue since Gradle is a preferred build tool that works just fine.

erictraut:
I think you are underestimating the impact this proposal would have on existing source bases. Many apps, RN extensions, and shared components make use of RN.WebView, and this change will break all of them. I strongly urge you to reconsider this.

satya164:
It's not going to happen without a deprecation path. And most the current `WebView` can be published as a separate legacy package so folks can switch to it with minimal changes when it's removed from the core.

satya164:
> Why not continue to support

That's exactly the problem. The core team doesn't support the WebView components and has other high priority tasks, so it's not going to change very soon. There are several bugs and missing features in the `WebView` component which have been there for a long time. Moving it to out of core means it can get more attention from the community. Would you really prefer an unmaintained module in the core than having a better-maintained version as a third-party package?

axe-fb:
Note that __This should NOT be a breaking change__. The way I am envisioning this happening is that react-native init would simply refer to the other module eventually. We are also working on a greater "slimmening" proposal (which I propose to write up here), so that we move many of the modules outside core.

The hope is that we reduce React Native's surface area and merge issues and PRs faster.

gwmccull:
@axe-fb if it's still referenced by `react-native init` will that help with bundle sizes at all?  It seems like one of the benefits of moving WebView outside the main repo would be to remove that code from bundles that don't use it

kelset:
Created a generic "overview" of the Slimmening discussion to create a context for this PR and avoid to go too "outside" the purpose of this PR https://github.com/react-native-community/discussions-and-proposals/issues/6

satya164:
Don't think `react-native init` referencing to the module can be considered a non-breaking change. If you're starting an app from scratch, there is nothing to break anyway. But it will be a breaking change for existing apps which are upgrading to a new version of `react-native`.

jamonholmgren:
> it will be a breaking change for existing apps which are upgrading to a new version of react-native

@satya164 is correct here, this will be a breaking change for existing apps that use WebView.

In another life, I built a [RubyMotion library](https://github.com/infinitered/ProMotion) that had an integrated maps and webview. I eventually moved maps into its own repo, but kept webview as a part of core because 1) it wasn't very big (same here), and 2) many people depended on it (same here).

However, the reality with React Native core is that the WebView as it currently stands is not sufficiently maintained, unlike my library's WebView.* So the overall goal is to reach a place where we have a WebView that is well maintained and reliable. And with the [upcoming deprecation of UIWebView](https://developer.apple.com/documentation/uikit/uiwebview) in iOS 12, the need has become more urgent.

> I think you are underestimating the impact this proposal would have on existing source bases. Many apps, RN extensions, and shared components make use of RN.WebView, and this change will break all of them. I strongly urge you to reconsider this.

I hear you @erictraut -- I'd like to see some examples so we can examine what the migration path looks like to those apps, extensions, and shared components. If it's as simple as adding a new import statement to the external legacy package, is that sufficient?

_* I don't blame Facebook for this at all -- they are providing an amazing community service by open sourcing React Native in the first place, and should be expected to focus on the things that mostly serve them. Just as Infinite Red focuses on the things, like Ignite, that we need._

fossecode:
Please make the extracted component compatible with Expo 🙏

grabbou:
I've been working on many apps recently that involve heavy-loaded interactions encapsulated inside a WebView and in most cases, I ended up using alternative packages provided by the community from superior performance.

I understand and agree with others that this is definitely going to be a breaking change, but at the same time, it's going to improve collaboration between different WebView projects out there, resulting in better performance for everyone. 

Once the package is extracted and fully tested to be working with `react-native link` and compatible with Expo at the same time, I think we are good to go.

anp:
@ide heads up that we should start evaluating what we're going to do with WebView.

proposals/0001-template.md
brentvatne: packages get handed off as companies/individuals either lose interest in a tool, close their doors, no longer want to maintain the package itself, the primary maintainer moves to another organization, or a handful of other reasons. this has downstream consequences - for example, if react-native-maps was scoped under `@airbnb` on npm, we'd be stuck with that or have to force consumers to change their imports and package.json just because of a change in project maintainer structure.

I understand your concern about wanting to get some return for the time you invest, but I think there are better ways we can do this than brand the package name itself.

jamonholmgren:
I understand @ide 's hesitation with connecting Expo (or `react-native init`) to a repo under Infinite Red's name. Point made and understood.

It's a significant, nontrivial undertaking to adopt a critical feature like WebView. We have 15 React Native developers at Infinite Red, but the majority of them are on client work, and only a few (like @GantMan and myself) are free to work on open source for a significant portion of our days.

So in order to justify taking on the amount of work that is likely to pop up (not only in working on the code base itself, but also support, answering issues, solving problems, etc), I'd like this to be under the Infinite Red banner. This gets our name out into the broader development community and drives consulting work our way. It makes it easy to allocate dev time to this project, because there's a clear ROI. If that goes away, it becomes sort of a side project for me, and not a very compelling one.

We often discuss in the community why open source projects feel unsustainable (even popular ones!) One solution is to have clear business support with reasonably clear benefits to doing so, and many React Native consultancies (including Callstack, Formidable, Geeky Ants, etc) participating in chunks of the ecosystem, helping each other and deriving value from it. Reasonable people might disagree with this approach. But it's the direction that Infinite Red tends to lean, and I think given our track record it's been a healthy one for the community as a whole.

My proposal:

1. Copy the current version of the WebView to a `react-native-community` repo and point any `react-native init`, Expo, etc to that one. Community supports it (I am willing to do the initial work to get it extracted and published there), but not Infinite Red.
2. Infinite Red publishes a fork of the project to our own repo, and we spend the time to get it healthy and solid and supported. Ignite and others can then use our WebView, and the community might rally around a more solidly maintained WebView.
3. At some point in time that @axe-fb decides, React Native deprecates the core version and eventually removes it from the repo (maybe with version 1.0.0? :) )

ide:
@jamonholmgren - That proposal would work for Expo. And I can relate to your sentiment around open-source contribution and needing for it to make sense even when you're putting your money where your mouth is and taking the lead to make WebView better.

(IMO it'd also work well if you wanted to contribute to `r-n-c/webview` and had an InfiniteRed sponsorship banner on the README, depending on how important the GitHub org is to you. That path might inspire another sponsor / agency to help out if you find that appealing and sustainable. But if you want "the InfiniteRed WebView" to be a thing that you provide and attracts clients, the second `infinitered/webview` repo + package make sense to me too.)

axe-fb:
@jamonholmgren - Needed some clarifications on the final proposal that you listed.

1. Whats the need to fork existing WebView to react-native-community, if it is not going to be maintained ? We may as well leave it the react-native main repo. Is it because you would like the use the exact same name in npm ?

2. Publishing the npm modules as @infinite-red/react-native-web-view will require developers to update their imports. Are you saying that when the infinite-red's version is official, you will also publish that as the node module that will be used by `require('react-native-webview')`. Will that work if we publish your fork to npm and remove the webview for RN core ?

grabbou:
Just my few cents here: We, at Callstack, also use WebView and would be
interested in helping with ongoing development when needed. I am sure there
might be other companies interested in doing so as well. I think the
solution where we create a repo under react-native-community and create a
'sponsors section' makes most sense to me. It leaves an open door to future
collaborators and sends out clear message that it's an all community
effort. I have found this indirect way of marketing the business behind
particular open source product very effective when it comes to gathering
leads for consulting work.

I think we need to figure out a good way of doing this. Having every single
module we strip out of the core under separate Github organisation, owned
by a different company will make it hell to control and maintain (my gut
feeling).

jamonholmgren:
@axe-fb 

> Whats the need to fork existing WebView to react-native-community, if it is not going to be maintained ? We may as well leave it the react-native main repo. Is it because you would like the use the exact same name in npm ?

Mainly to counteract the "power of the default", where most people will want to use the built-in version rather than a third party, just because it "seems" like it would be better supported. Having no WebView in core means that third parties could be evaluated on their merits.

> Publishing the npm modules as @infinite-red/react-native-web-view will require developers to update their imports. Are you saying that when the infinite-red's version is official, you will also publish that as the node module that will be used by require('react-native-webview'). Will that work if we publish your fork to npm and remove the webview for RN core ?

Regardless of where we publish it (even `react-native-webview`), developers will have to update their imports. Currently you import this way:

```jsx
import { WebView } from 'react-native'
```

With any third-party published module, even under the community, you wouldn't be able to import from React Native:

```jsx
import { WebView } from 'react-native-webview'
```

So the sense I'm getting is that everyone would prefer the main WebView repo to be under `react-native-community`. I'd at least like some sort of Infinite Red sponsorship link, but otherwise it's okay with me. If @grabbou's Callstack wants to contribute as well, that's cool.

ide:
> most people will want to use the built-in version rather than a third party, just because it "seems" like it would be better supported

To add this, several components in the react-native repo currently receive very minimal support and have several PRs open. Even though some of those PRs probably don't make sense to take e.g. they aren't maintainable or are overfitted or underfitted solutions, there's a mismatch between how supported people think the react-native repo's components are and how much the components are actually supported and guided by a steward with a vision.

That's why I think it's healthier for the RN project, its community, and Facebook for the react-native repo to focus on the core (bridge, scheduler, UIManager, Yoga, etc...) -- this is where the highest-impact commits from Facebook have been -- and to set up clear stewardship with capable contributors like Infinite Red and, ultimately, clear expectations with the open-source community that the lines between repos are aligned with the lines of responsibility for the different parts of the code that currently lives in react-native today.

proposals/0001-template.md
jamonholmgren: @brentvatne Thanks for the feedback. I've removed the namespace and went with just `react-native-webview`.

jamonholmgren:
I've updated the proposal, removing the NPM namespace. cc @brentvatne @axe-fb @ide 

Perhaps the Github repo could still live in `infinitered/react-native-webview` but not the NPM module? (I'm just thinking out loud here; open to feedback.)

proposals/0001-webview.md
```diff
@@ -0,0 +1,64 @@
...
+
+It also means there won't be an officially supported WebView in React Native, which may feel to some to be missing key functionality.
+
+## Alternatives
+
+The primary alternative is to keep the WebView in core. There are also existing third party WebViews, although none that are reliably cross-platform.
+
+## Adoption strategy
+
+This is a breaking change and existing React Native authors who use WebView will need to update their apps to access one of the third party solutions. For most, it should be as simple as `yarn add @infinitered/react-native-webview` and update their import statements.
```
ide: need to update this too

grabbou:
@jamonholmgren to be honest, I am still concerned about the ownership of the code, that is, the location where it lives. I would much prefer the project to live under `react-native-community` as it has been stripped out from `react-native` where it belonged for a while. I described it in details in my [previous comment](https://github.com/react-native-community/discussions-and-proposals/pull/3#issuecomment-410467165). Would love to hear your thoughts about it. Similar approach has been followed by the Hapi.js community, where each module that builds the framework (e.g. Joi) has a separate repository with lead developer assigned and sponsors section. From the very first moment we have created this organisation, my goal was to have a community that works together on most popular React Native modules that it uses. This is nothing new and in fact, other communities, like ReasonML have adopted this from the very beginning, via their `reasonml-community` organisation.

After React Native EU, I was planning to submit proposal of moving CLI from the core repository. I think we should be consistent here as far as the process of making the core thinner and make no precedences as to where the project ends up being maintained.

proposals/0001-webview.md
jamonholmgren: Oops, yes.

jamonholmgren:
@grabbou Yes, I'm okay with putting it on the `react-native-community` Github org if that's what the general consensus is, but a sponsorship listing on the repo isn't enough for me to devote lots of Infinite Red engineering time to it. I'm willing to spend some of my own personal open source time helping out on it though, and can publish the first version for sure (in progress right now).

(My personal opinion is that maintainability of open source is a problem when nobody "owns" it. When a company owns it and has a business reason to support it, it's going to be much better maintained, IMO. I'd be happy to be proven wrong, but that's been my experience with community open source.)

grabbou:
If that doesn't work for you, maybe we should think of a model that suits
us all. For instance, maybe a section on the website or a rrschout from
Twitter account would do.

What I am really trying to pursue here is a model where there's healthy non
profit community of RN developers working together. I'd love to have
Infinite Red on board since we all use those modules to get our business
done. This won't work without us being involved.

hramos:
FYI, a bunch of WebView changes just landed in master. We're writing up what these changes will mean for WebView users. It'll probably become a blog post. We'll link back to this proposal in any case.

fungilation:
Finally, WKWebView!

I'm stuck using abandoned https://github.com/alinz/react-native-webview-bridge due to this: https://github.com/facebook/react-native/pull/9762#issuecomment-260583009. Would be great to allow setting `window.postMessage` to override any existing one in an arbitrary webpage, or avoid name collision by moving RN's injection to `window.rnPostMessage`. Had a long discussion before on [#9762](https://github.com/facebook/react-native/pull/9762#issuecomment-260583009) and don't remember where else on this and it didn't go anywhere.

Alongside or after the work on adding WKWebView, which as I understand it has new APIs which makes implementing this easier? I'm willing to investigate and implement if lack of interest is the issue with having a functional, non-colliding `window.postMessage`. I just need direction on how to approach getting this into RN, if it's wanted. As I need this in my app, and I want to move off unmaintained `react-native-webview-bridge`.

axe-fb:
@jamonholmgren  - What are the next steps on this ? Would you still want to lead this effort ?

axe-fb:
@fungilation  - regarding postMessage - we have not made the change since it was a breaking API. However, now that we are spinning it out of core, breaking changes are fair game. Once this is moved out into the community, we can continue the discussion on the repo.

fungilation:
Thanks @axe-fb. Agreed with order of things. Glad to hear that spinning WebView out of core means more breaking change but necessary improvement like `postMessage` can happen. Especially since RN<>WebView message passing/RPC can now happen natively without hacks with WKWebView.

jamonholmgren:
Apologies for being out of contact on this. I've been way more busy with work than I expected. Big kudos to the Facebook team for landing this work!

@axe-fb I'm still interested in working on spinning this out of core, depending on how quickly you want it done. I have some work partially done that I can adapt to this, and then push to a new repo on `react-native-community`. I realize it's been a couple weeks here, so if anything has changed in the meantime, feel free to ping me on Slack.

jamonholmgren:
One comment I have is that it seems that the extraction could go all-in on WKWebView and leave behind the legacy UIWebView code. Right now, [there is code](https://github.com/facebook/react-native/blob/master/Libraries/Components/WebView/WebView.ios.js#L471-L475) that switches between the two view managers. I would suggest that we drop that in the extracted code base and only support WKWebView. Any thoughts?

fungilation:
I'm all for going with WKWebView. Performance, maintained by Apple, not deprecated and all that. But perhaps have a phase out period (after spinout) so it's more battle tested before dropping UIWebView code?

jamonholmgren:
@fungilation I have UIWebView and WKWebView both working in my third-party package, now. It makes sense to support them both for at least a short while.

kelset:
Awesome work @jamonholmgren , we'll make sure to link it into the Changelog :)

satya164:
> No Flow or TypeScript, yet. Any preferences as to which?

I think Flow is a better choice because RN core uses Flow. But I think ideally we should have both.

fungilation:
Does what RN uses matter for WebView after it is spun out? If not, TypeScript have my vote.

Titozzz:
I'm using typescript too but better solution would be to support both I guess

jamonholmgren:
I've migrated (with @Titozzz's help) the issues labeled `WebView` over to https://github.com/react-native-community/react-native-webview/issues. If the RN core team is okay with moving forward with this proposal, all of the WebView-related issues on the React Native repo can be closed, and going forward new WebView-related issues can be posted to the new repo.

We plan to go through pull requests at a later date.

koenpunt:
> 1) There are a lot of small useful modules which uses WebView behind the scenes for things to work. Some quick examples are react-native-qrcode, react-native-webview-crypto, react-native-editor, etc...

I think the solution for these implementations is to not use a webview (e.g., the QR code thing can also be achieved using React ART), or include a custom implementation of a native webview themselves.

axe-fb:
Merging this PR, since this is now approved. Thanks a lot @jamonholmgren  for the amazing work !!

webfangjun:
react-native-webview source can't load resource (URL:file://) in the iOS13. but URL = http is OK.

proposals/0001-template.md
licguzman: Hello i'm new i made a  webview app using this code

licguzman: import React, { Component } from 'react';
import WebView from 'react-native-webview';

export default class App extends Component {
  render() {
    return (
      <WebView
        source={{ uri: 'http://degrapack.com/radical' }}
        style={{ marginTop: 20 }}
      />
    );
  }
}

licguzman: import React, { Component } from 'react';
import WebView from 'react-native-webview';

export default class App extends Component {
  render() {
    return (
      <WebView
        source={{ uri: 'http://google.com' }}
        style={{ marginTop: 20 }}
      />
    );
  }
}

licguzman: but i want to go back in the webview app and the app closes, can you share the code I Need to fix it

sopu0000:
How to use camera through Webview? can anyone help ? it always says permission denied.

DakshAg62:
I am using react native webview to render a site in my app. My problem is that if I click on add to cart button it disappears on adding to cart but doesn’t come back if item is removed (like on website). but on website it comes back on removing the item form the cart. How can i solve this issue?

Scutua (approved):

Issue #7, opened 2018-08-06
Debugging JSC directly


axe-fb:
# What
The ability to attach to JSC directly to debug React Native JavasScript code. 

# Background
Currently, we use Chrome to run the JS code when React Native is in the debug mode. When we start the debug mode, a new web socket connection is established via the packager to an instance of Chrome. We then use Chrome developer tools to debug the JS code. Similar mechanisms are used in the VSCode debugger for React Native, and the Idea react native plugin.

# Why
When Native Modules today use a synchronous API, Chrome debugging fails. Sync API for native modules are not common, and at many places, such module provide alternative async calls. 
With Fabric, there would be many more cases that utilize the synchronous API. Fabric will also use the new JavaScript Interface (called JSI), which lets JavaScript hold references to C++ objects. 
Both these cases may not debuggable with the current setup. 

# How
There are currently two possible approach to support debugging in the future for OSS React Native projects. 

1. Support Synchronous debugging in Chrome. For this, whenever we move over from WebSockets to Sync-XHR whenever a synchronous API is encountered. This could be slow, and may not support the concept of Host Objects.

2. Use Node for Synchronous debugging - Currently VSCode uses this approach. Instead of starting Chrome, it starts up Node, and debugs node using the node debug protocol. Node has support for Synchronous HTTP/XHR, so this could be leveraged. We still need to see how JS holding references to C++ objects would work.

3. Debug JSC directly on the device - this seems to be the best approach, but may also require the maximum amount of work. Currently, Safari allows JSC on the device to be debugged directly. The JSC that we use in Android is pretty old and may not support debugging. We would need to find a good debugging story. 

Side Note - We would also need to start supporting debugging Chakra in case of RN-Windows.

gwmccull:
I think the debugging in Node would be cool.  I'm guessing that would mean you could use the Node in an Electron app (ie, React Devtools) to do the debugging which could allow developers to build some pretty powerful debugging interfaces

axe-fb:
Not really use Node in electron. I just mean that instead of JSC, the JS code will run on Node, and debuggers could be attached to Node.

sahrens:
Would it really be "Node"? Wouldn't it be more like just using V8 directly as the VM, like NativeScript does? "Node" implies a whole set of runtime environment and builtins that we wouldn't want.

sahrens:
But I also don't see a great reason to use V8 instead of Chakra if Microsoft is going to get Chakra working anyway for windows.

[...]

Issue #4, opened 2018-07-31
React Native Fabric (UI-Layer Re-architecture)


kelset:
### Intro

With this issue I'd like to try and create a "one stop" for all the information available around the future re-architecture of the UI-Layer of React Native, codenamed "Fabric".

### Terminology
* JSI: JavaScript Interface, it's a unified lightweight general purpose API for (theoretically) any JavaScript virtual machine. It enables every other piece of the rearchitecture. ([dedicated issue](https://github.com/react-native-community/discussions-and-proposals/issues/91))
* TurboModules: re-architecture of NativeModules, also using JSI. ([dedicated issue](https://github.com/react-native-community/discussions-and-proposals/issues/40))
* CodeGen: a tool to "automate" the compatibility between JS and native side. ([dedicated issue](https://github.com/react-native-community/discussions-and-proposals/issues/92))

### TL;DR

From @axe-fb's [blogpost](http://blog.nparashuram.com/2019/01/react-natives-new-architecture-glossary.html), here's a temporary description of Fabric (*please consider that this is not yet finalized, it may change in the future*)

> In the current architecture, all UI operations (like creating native views, managing children, etc). are handled by a native module called UIManagerModule. The React Reconciller sends UI commands over the bridge, which are eventually handled by this module and delegated to UIImplementation. This in turn creates shadow nodes that represent the layout tree and are passed to Yoga to determine the relative co-ordinates based on the Flex box styles that are passed in from JS. 
In the new system, the UI operations are directly exposed to JavaScript as functions using the JSI interface described above. The new UI manager can then create ComponentDescriptors and the Shadow Nodes for specific view types (like Text, View or Images), and then communicate with Java/ObjC to draw platform specific UI.

### Available Materials

It was first announced in the ["State of React Native 2018"](http://facebook.github.io/react-native/blog/2018/06/14/state-of-react-native-2018) blogpost by @sophiebits.

In July, @hramos linked a broadcast of a tech talk that happened in FB, which covered the ongoing work on the new React Native architecture codenamed "Fabric": facebook.com/hramos/videos/10101317533036249 (go to 8:18 for the presentation start - audio is a bit bad)

Further down the line, it was presented at ChainReact Conf 2018 by @axe-fb in the talk ["The State of React Native"](https://www.youtube.com/watch?v=83ffAY-CmL4).

[...]

PR #11, opened 2018-08-09
Proposal about 0002-Turbo-Modules.md

axe-fb:


kelset:
I'm a bit confused, but I understand this is a WIP so I guess once it's complete it will be clear.
Just one question:

> This proposal discusses a new architecture to speed up the way Native Modules are initialized and invoked from JavaScript.

Is this something else from Fabric and from the JSI? Or, like, a proposal that aims to explain a side of the Fabric approach?

Just asking because if it's "Fabric related" I think that adding a new label for Fabric could help.

axemclion:
This is not related to fabric, but is built on to of JSI. It could have been done independent of JSI too, but taking advantage of using host objects makes this better.

While fabric is mostly for view managers, this is for native modules.

kelset:
So, basically, if I understand correctly:

```
Native <--> JSI <--> Turbo Module <--> Fabric <--> [code you actually write]
```

Is this rough scheme correct?

axe-fb:
Not really. While UIManagerModule today is a Native Module, it does not have to be one. We are looking to move FabricUIManager to directly be able to install bindings on JSI. This way, the diagram would be 

Java View Managers <--> JNI <--> C++ Shadow Tree <--> JS code
Java Native Modules <--> JNI <--> C++ <--> JSI <--> JS Code

kelset:
Ah wow, that's a great diagram - I think it would be helpful to have inside the proposal md :)

grabbou:
Interesting - that sounds like what we have talked about two weeks ago at Facebook. I thought that it was part of Fabric when I heard about it - sounds really interesting. Looking forward to the updated proposal that goes a bit more technical into how the native modules work right now.

I am currently planning on working on better `link` and this is highly interesting to me.

karanjthakkar:
@axe-fb Possibly silly question: Since the links in the RFC point to Java code and [this diagram](https://github.com/react-native-community/discussions-and-proposals/pull/11#issuecomment-411823611) also talks about the lifecycle of native modules in Java land, I was wondering if the optimisations mentioned in this RFC also apply to the iOS side of things?

fkgozali:
> I was wondering if the optimisations mentioned in this RFC also apply to the iOS side of things?

It’ll be the same concept in iOS, just without JNI, and it’ll beObjC classes instead of Java classes. The C++/JSI layer will be shared across platforms.

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+
+2. There is no simple way to check if the Native Modules that JavaScript calls are actually included in the native app. With over the air updates, there is no easy way to check if a newer version of JavaScript calls the right method with the correct set of arguments in Native Module.
```
eliperkins: > if a newer version of JavaScript

What do you think about wording this "if a newer version of the JavaScript bundle" so that folks don't confuse it with a newer JavaScript version like ES2030 or whatever 😄

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+
+2. There is no simple way to check if the Native Modules that JavaScript calls are actually included in the native app. With over the air updates, there is no easy way to check if a newer version of JavaScript calls the right method with the correct set of arguments in Native Module. 
+
+3. Native Modules are always singleton and their lifecycle is typically tied to the lifetime of the bridge. This issue is compounded in brownfield apps where the React Native bridge may start and shut down multiple times.
```
eliperkins: s/singleton/singletons

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+
+4. During the startup process, Native Modules are typically specified in [multiple packages](https://github.com/facebook/react-native/blob/b938cd524a20c239a5d67e4a1150cd19e00e45ba/ReactAndroid/src/main/java/com/facebook/react/CompositeReactPackage.java). We then [iterate](https://github.com/facebook/react-native/blob/407e033b34b6afa0ea96ed72f16cd164d572e911/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java#L1132) over the list [multiple times](https://github.com/facebook/react-native/blob/617e25d9b5cb10cfc6842eca62ff22d39eefcf7b/ReactAndroid/src/main/java/com/facebook/react/bridge/NativeModuleRegistry.java#L46) before we finally give the bridge a [list of Native Modules](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/CatalystInstanceImpl.java#L124). This does not need to happen at runtime.
```
eliperkins: This seems to be specific to Android. Is it worth mentioning the process that React Native for iOS uses here?

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+
+5. The actual methods and constants of a Native Module are [computed during runtime](https://github.com/facebook/react-native/blob/master/ReactCommon/cxxreact/ModuleRegistry.cpp#L81) using [reflection](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/JavaModuleWrapper.java#L75). This can also be done using build time. The individual method calls themselves are sent over a message queue. 
+
+## Background
+
+### JavaScript Interface
+The communication between JavaScript and the VM is being standardized using a JavaScript Interface (__JSI__). [Fabric](https://github.com/react-native-community/discussions-and-proposals/issues/4) uses JSI to [expose methods](https://github.com/facebook/react-native/blob/5d9326be29be8688f2238c72c18a9037f983c77d/ReactAndroid/src/main/java/com/facebook/react/fabric/jsc/jni/FabricJSCBinding.cpp#L264) to JavaScript code. Using JSI, JavaScript can hold reference to C++ Host Objects and invoke methods on them.
```
eliperkins: s/reference/references

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+### JavaScript Interface
+The communication between JavaScript and the VM is being standardized using a JavaScript Interface (__JSI__). [Fabric](https://github.com/react-native-community/discussions-and-proposals/issues/4) uses JSI to [expose methods](https://github.com/facebook/react-native/blob/5d9326be29be8688f2238c72c18a9037f983c77d/ReactAndroid/src/main/java/com/facebook/react/fabric/jsc/jni/FabricJSCBinding.cpp#L264) to JavaScript code. Using JSI, JavaScript can hold reference to C++ Host Objects and invoke methods on them. 
+
+### Native Modules 
+ > [TODO] - A detailed section on how native modules work today. 
+
+## Detailed design
+
+## Defining Native Modules
+Internally, we use flow types to define the Native Modules and their method and signatures (also called "shape of Native Modules"). This is then used to generate Java and ObjC classes with method stubs with matching signatures, which can be extended for the actual implementation. We plan to now generate C++ classes with the same shape, that can then be used to call into the Java/ObjC implementations.
```
eliperkins: Does "internally" here refer to Facebook or React Native in general?

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+### Native Modules 
+ > [TODO] - A detailed section on how native modules work today. 
+
+## Detailed design
+
+## Defining Native Modules
+Internally, we use flow types to define the Native Modules and their method and signatures (also called "shape of Native Modules"). This is then used to generate Java and ObjC classes with method stubs with matching signatures, which can be extended for the actual implementation. We plan to now generate C++ classes with the same shape, that can then be used to call into the Java/ObjC implementations. 
+
+## Initial Processing
+All initial processing will be removed as C++ now knows the shape of Native Modules. A proxy similar to the [NativeModuleProxy](https://github.com/facebook/react-native/blob/da7873563bff945086a70306cc25fa4c048bb84b/ReactCommon/cxxreact/JSCExecutor.cpp#L141) will be defined in C++ and [exposed in JavaScript](https://github.com/facebook/react-native/blob/a93e281428f47ba355ba79cda3e98ca7da677636/Libraries/BatchedBridge/NativeModules.js#L152). The initialization can typically happen when the RootView is initialized, but once we are independent of the bridge, this can also happen at other times.
```
eliperkins: This will still require a compilation step, however. Is that correct?

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Initial Processing
+All initial processing will be removed as C++ now knows the shape of Native Modules. A proxy similar to the [NativeModuleProxy](https://github.com/facebook/react-native/blob/da7873563bff945086a70306cc25fa4c048bb84b/ReactCommon/cxxreact/JSCExecutor.cpp#L141) will be defined in C++ and [exposed in JavaScript](https://github.com/facebook/react-native/blob/a93e281428f47ba355ba79cda3e98ca7da677636/Libraries/BatchedBridge/NativeModules.js#L152). The initialization can typically happen when the RootView is initialized, but once we are independent of the bridge, this can also happen at other times. 
+
+## Invoking methods on Native Module
+When the JavaScript code like `require('NativeModule').DeviceInfoModule` is called, the getter on the proxy is invoked. It would return a reference to object generated in C++ with the shape of the Native Module. Method calls on this reference would simply call over to Java or ObjC. The actual call may also need to have checks to ensure the type and count of arguments.
```
eliperkins: Would you mind removing "simply" here? While it may be simple for you, using this word here might make others feel stupid if they don't understand what's so simple about this step.

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Invoking methods on Native Module
+When the JavaScript code like `require('NativeModule').DeviceInfoModule` is called, the getter on the proxy is invoked. It would return a reference to object generated in C++ with the shape of the Native Module. Method calls on this reference would simply call over to Java or ObjC. The actual call may also need to have checks to ensure the type and count of arguments. 
+
+> [TODO] - A work in process.
+
+## Open Questions
+1. How would the C++ object corresponding to the shape of a Native Module know the Java class (for example com.facebook.react.nativemodule.DeviceInfo) that it needs to call. Does this need to be a switch/case or a map in Java/ObjC that is constructed manually ? Would this be done using BUCK dependencies ? 
+2. Today, native modules usually take `ReactApplicationContext` as a parameter when they are constructed, but they can also be constructed in other ways like using a builder, or multiple parameters in the constructor. How would this be enforced ? 
+3. How would the C++ code gen happen in Open Source, where BUCK is not used ?
```
eliperkins: Perhaps we should look into a "compiler" that is similar to the Relay compiler. That is, some tool that would do the code gen for us in OSS, that BUCK could then depend on.

In the OSS version, we could add this as a build script to projects outside of BUCK.

proposals/0002-Turbo-Modules.md
fkgozali: Facebook

proposals/0002-Turbo-Modules.md
fkgozali: yeah, everything still needs to be compiled, but c++ doesn’t need to ask Java/ObjC about the “shape of the module” anymore, reducing cost of starting up the system

proposals/0002-Turbo-Modules.md
fkgozali: yeah, we can treat this as a “script” that the build system calls/executes, e.g. buck calls this script for us internally at Facebook

proposals/0002-Turbo-Modules.md
fkgozali: yeah probably good to have section explaining iOS, though the end goal is to share the c++/jsi layer across platforms

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
+# RFC0002: Turbo Modules ™ 
+_Its blazing fast_
+
+## Summary
+
+React Native uses [Native Modules](https://facebook.github.io/react-native/docs/native-modules-ios) as a way for JavaScript  to invoke functions in Java/ObjC. Listed under the API section in the [React Native documentation](https://facebook.github.io/react-native/docs/getting-started), popular native modules include Geolocation, Device Information, Async Storage, etc.
```
shergin: I would explicitly mention here that Native Modules does *not* directly tight with UI rendering pipeline and that currently-existing View Managers (which are also Native Modules) will go away soon.

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+
+5. The actual methods and constants of a Native Module are [computed during runtime](https://github.com/facebook/react-native/blob/master/ReactCommon/cxxreact/ModuleRegistry.cpp#L81) using [reflection](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/JavaModuleWrapper.java#L75). This can also be done using build time. The individual method calls themselves are sent over a message queue. 
+
```
shergin: I would also mention here quickness/grayish-area of methods returning result synchronously, I feel we have to "legalize" them.

proposals/0002-Turbo-Modules.md
shergin: I don't think that initialization of a native module should/can be tight with life-cycle of RootView; it can be before or after depending on particular module.

proposals/0002-Turbo-Modules.md
```diff
@@ -0,0 +1,53 @@
...
+> [TODO] - A work in process.
+
+## Open Questions
+1. How would the C++ object corresponding to the shape of a Native Module know the Java class (for example com.facebook.react.nativemodule.DeviceInfo) that it needs to call. Does this need to be a switch/case or a map in Java/ObjC that is constructed manually ? Would this be done using BUCK dependencies ? 
+2. Today, native modules usually take `ReactApplicationContext` as a parameter when they are constructed, but they can also be constructed in other ways like using a builder, or multiple parameters in the constructor. How would this be enforced ? 
+3. How would the C++ code gen happen in Open Source, where BUCK is not used ? 
+4. Since the generated C++ host objects have the same lifetime of the JS scopes they are referred in, what is the API to tell Java/ObjC when these objects are deallocated ?
```
shergin: Why do they have to have same lifetime as JS scope especially if they are not singletons?

proposals/0002-Turbo-Modules.md
fkgozali: existing NativeModules system already supports returning values synchronously - not a gray area

proposals/0002-Turbo-Modules.md
fkgozali: The C++ host objects have destructors (which will be called when the JS object gets GC'ed) which can tell platform specific classes to "clean up" or deallocate themselves.

proposals/0002-Turbo-Modules.md
axe-fb: I agree - all I am saying is that it MAY happen at the rootView initialization for now, but this could also happen at other times - typically during a "bridge-prewarmup"

proposals/0002-Turbo-Modules.md
karanjthakkar: @fkgozali Is there an officially supported way to return values from Native Modules synchronously? Or is this just something that's present internally but not exposed?

fkgozali: The support has existed for a while, you may use it, but just note that Chrome debugger doesn't support such synchronous call, so it may make debugging JS harder for you. It's kinda why it's not fully documented. See https://github.com/react-native-community/discussions-and-proposals/issues/7 for some context on debugging.

iOS: Use `RCT_EXPORT_SYNCHRONOUS_TYPED_METHOD ()` instead of `RCT_EXPORT_METHOD()`: https://github.com/facebook/react-native/blob/ca898f4367083e0943603521a41c48dec403e6c9/React/Base/RCTBridgeModule.h#L175

Android: add `(isBlockingSynchronousMethod = true)` to your `@ReactMethod` annotation:
https://github.com/facebook/react-native/blob/c8e000b19a3dc47d58bc3ea1ede8211a370d59bb/ReactAndroid/src/test/java/com/facebook/react/bridge/BaseJavaModuleTest.java#L103

proposals/0002-Turbo-Modules.md
shergin: Oh, I know, I know. My point is that JS scope usually can increment and decrement ref-counter, this does not always mean that the life-times are identical. I think we are on the same page, I would just clarify the spec.

dryganets:
It sounds like we can't use typescript for react-native modules in new architecture unless we will generate/define the flow types for every module.

Is there a way to avoid flow usage?

acoates-ms:
I believe the code gen tool has an intermediate data format for defining the module shape.  Facebook happens to be generating that from flow, but an additional generator from typescript could easily be written and still use the same code gen. We can probably also add code gen that generates d.ts files also, which would allow the typescript definitions files to always be up to date for native modules -- even ones defined in flow, which would be an improvement over the hand generated ones people currently have to use.

proposals/0002-Turbo-Modules.md
aleclarson: @shergin Is there more information on the coming removal of View Managers?

fkgozali: view managers will be handled as part of fabric rollout: https://github.com/react-native-community/discussions-and-proposals/issues/4

exact details for migration path is TBD

proposals/0002-Turbo-Modules.md
jamonholmgren: Could use "directly".

axe-fb:
I think we should merge this proposal PR into the repo. Like Fabric, we can use an issue for addressing questions. As we have updates, we can continue to update this doc. 

If no one says otherwise, I plan to merge this on Sep 11, 2018, 10 am PDT.

radex:
> This is then used to generate Java and ObjC classes with method stubs with matching signatures, which can be extended for the actual implementation.

I'm not sure I'm understanding this 100% right, so: Are the generated ObjC classes just to make it easier to the user, or does the new system still rely on macros or ObjC runtime somehow to figure out the native module's methods and arguments? (I'm guessing no.)

In other words: will I be able to write plain Swift native modules, without having to hand-write the ObjC `.m` stub?

fkgozali:
> Are the generated ObjC classes just to make it easier to the user, or does the new system still rely on macros or ObjC runtime somehow to figure out the native module's methods and arguments?

What's generated is just objc `@protocol` and Java interface (or abstract class) that your module class just need to implement. The methods in those generated output are based on JS type (e.g. Flow), so you won't need any of the `RCT_EXPORT_*` macros or `@ReactMethod` annotations to expose your method signature to JS. This also provides build-time check to make sure that the JS and native "signature" are the same.

For Swift, in theory we could improve the scripting/codegen tool to additionally output Swift protocols.

radex:
Thanks for clarifying! Unless it’s much more complicated than I expect, I’d be happy to contribute Swift codegen 👍

proposals/0002-Turbo-Modules.md
matthargett: just to note: both companies I have worked at ended up using sync calls in a surgical way for notable user-visible perf improvements.

proposals/0002-Turbo-Modules.md
matthargett: what about modules that are 100% implemented in C++? we have a few modules like this, as did my previous employer. my guess is that it would be an even more efficient communication: JS<->C++<->JS, instead of JS<->C++<->Java/ObjC<->C++<->JS, right?

matthargett: to this end, I think a sample AsyncStorage module that uses C++ iostreams to demonstrate the happy and error paths would be useful. specifically, an interface that takes this kind of race/lifecycle issue into account: https://github.com/Microsoft/react-native-windows/pull/1888

fkgozali: C++ only modules are definitely a good use case to cover. We’ll have some hooks after the objc/java hooks are finalized. We do have some C++ based modules internally, but all of them today need to be backed by objc/java, via CxxNativeModule (eg deferring to the os API to get storage path, etc).

For now, first step is to make TurboModule compatible with CxxNativeModule mechanism, then improve the binding to allow direct C++ impl without objc/java as the second step.

axe-fb:
I would like to merge this in since we are working on this.

fkgozali:
@axe-fb yup let’s do that, then we can open an issue for q&a like fabric

proposals/0002-Turbo-Modules.md
matthargett: great! avoiding the cost of even "fast" marshaling across JNI/.NET is key for scraping back CPU cycles and memory usage, especially in case of AsyncStorage, WebSocket, etc where giant blobs are potentially being copied today.

proposals/0002-Turbo-Modules.md
matthargett: keeping a "warm" bridge around initialized with the common first slice of the RAM bundle was used in a few RN Windows apps depending on if people launched the full app, or the "mini" tray app. I suspect this optimization is more common than even I know.

matthargett:
so... merge?

Issue #15, opened 2018-08-11
Use react-native as a dynamic framework on iOS


dehlen:
Hi,

First and foremost I'd like to thank you and all the OS contributors for the effort going into this project and enabling so many people to write performant native applications with a single code base and with a great tooling. Me and my team love to hack with react-native.

Currently we use Carthage extensively at my team for dependency management on iOS. I wondered whether it would be possible to use react-native with Carthage as a dynamic framework. This is because
a) we don't want to mix and match multiple dependency management solutions
b) we could cache react-native as a dynamic framework. This way the dependency with all of it third-party dependencies could build one time and after that the CI server could use the cached version.

I think the approach of using react-native as a dynamic framework would increase the build time of our application by a huge factor. Also currently the whole react-native dependency is located in a dynamic framework of ourselves. This framework is then added to an app target which forces us to link the whole Xcode project of our framework and build the sources on the app build because we can't just integrate our framework via Carthage. This is mainly due to the fact that the dynamic framework of ours would need to export all the relevant react-native headers in order to import the framework correctly.

I would love to hear your thoughts whether
a) Carthage support is planned
b) if the support for Carthage is not on your list how we should deal with the current dependency problematic. We currently have to rebuild the whole react-native Xcode project for every app build on our CI server. This leads to huge increases of our build time and prevents us from benefiting the fast iteration cycles react-native could offer.

Also if no support for Carthage is planned maybe someone could help out with a sample project or script which includes the static react-native library and exports the headers correctly so we could maintain a dynamic wrapper framework to achieve our goals ?

Since react-native dropped support for iOS8 in 0.56.0 I guess providing a dynamic framework for iOS has no real disadvantage because since iOS8 dynamic frameworks are supported and could help a lot of people as other issues already have stated.

I am happy to hear your answers and to start a discussion about this.

eliperkins:
I think the only anecdotal evidence I have of why React Native currently doesn't support vendoring a dynamic framework is that Facebook itself uses static libraries for everything.

I wonder if @alloy has any context into this, as someone who has spent a large amount of time setting up React Native with CocoaPods. 🤔

> I guess providing a dynamic framework for iOS has no real disadvantage

Currently, even with the latest iOS versions, loading dynamic frameworks are still slower than loading static libraries. There's (lots) more context here: https://github.com/artsy/eigen/issues/586

shergin:
Eli @eliperkins,

> Facebook itself uses static libraries for everything.

[...]

PR #18, opened 2018-08-15
RFC for Improving CocoaPods support

orta:
This should make CocoaPods support match the Xcode projects currently used, and provides a technique to make sure that all testing can be done offline,

### DIFF
--- a/proposals/0003-cocoapods-support-improvements.md
+++ b/proposals/0003-cocoapods-support-improvements.md
@@ -1,0 +1,117 @@
+---
+title: CocoaPods Support Improvements
+author:
+- Orta Therox
+- Eloy Durán
+date: 2018-08-15
+---
+
+# RFC0003: CocoaPods Support Improvements
+
+## Summary
+
+Simplify CocoaPods support in React Native, by reducing the differences between how CocoaPods maps React Native into the existing set of Xcode projects.
+
+## Basic example
+
+CocoaPods support in React Native is a tad tricky, because it’s trying to map a complex set of Xcode projects into a single dependency. The Podspec for React Native does a lot of work under the hood at ~300 LOC. This is, in part, because we use a CocoaPods abstraction called subspecs to consolidate out the React Native library into one single library.
+
+If we split the `React.podspec` into many podspecs, one per Xcode project. Roughly ~10ish Podspecs per React Native release.  These would match up one-to-one to the Xcode projects so that when CocoaPods support is broken in React Native, an error report would make it obvious which part of the project is broken.
+
+This means we can also move React Native and we can get CocoaPods support back in CI. This route allows CocoaPods support to be tested without network access (all the required information would be inside the same)
+
+The CocoaPods React dependency would then depend on React-ART, and React-RCTSettings etc.
+
+## Motivation
+
+For teams adding React Native to existing apps CocoaPods support let's RN be treated as a dependency like all of the other ones. CocoaPods is built with a lot of escape valves for consumers to fix issues at runtime (you see people passing around `post_install` [fixes](https://github.com/facebook/react-native/issues/20182#issuecomment-409131862) for example, and I built a [CocoaPods plugin](https://github.com/orta/cocoapods-fix-react-native) to try consolidate them all)
+
+I think one of the a key reasons for the drift between the CocoaPods support and the Xcode projects is because they're not mapped together.
+
+## Detailed design
+
+```sh
+rm React.podspec
+
+touch React/React.podspec
+pod spec lint React/React.podspec
+
+touch Libraries/ART/React-ART.podspec
+touch RNTester/React-RNTester.podspec
+touch Libraries/Blob/React-RCTBlob.podspec
+# etc
+```
+
+Implementing that is reasonably straight forwards. Then we have two ideas about how the Podspecs can be distributed.
+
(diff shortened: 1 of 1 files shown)


proposals/0003-cocoapods-support-improvements.md
```diff
@@ -0,0 +1,117 @@
...
+
+This means we can also move React Native and we can get CocoaPods support back in CI. This route allows CocoaPods support to be tested without network access (all the required information would be inside the same)
+
+The CocoaPods React dependency would then depend on React-ART, and React-RCTSettings etc.
+
+## Motivation
+
+For teams adding React Native to existing apps CocoaPods support let's RN be treated as a dependency like all of the other ones. CocoaPods is built with a lot of escape valves for consumers to fix issues at runtime (you see people passing around `post_install` [fixes](https://github.com/facebook/react-native/issues/20182#issuecomment-409131862) for example, and I built a [CocoaPods plugin](https://github.com/orta/cocoapods-fix-react-native) to try consolidate them all)
+
+I think one of the a key reasons for the drift between the CocoaPods support and the Xcode projects is because they're not mapped together.
```
eliperkins: What implications does this have to the existing consumers of the Xcode projects? My understanding is that they'll be unused by CocoaPods. How will this ensure that CocoaPods and the existing Xcode projects stay in sync?

proposals/0003-cocoapods-support-improvements.md
```diff
@@ -0,0 +1,117 @@
...
+
+If we split the `React.podspec` into many podspecs, one per Xcode project. Roughly ~10ish Podspecs per React Native release.  These would match up one-to-one to the Xcode projects so that when CocoaPods support is broken in React Native, an error report would make it obvious which part of the project is broken.
+
+This means we can also move React Native and we can get CocoaPods support back in CI. This route allows CocoaPods support to be tested without network access (all the required information would be inside the same)
+
+The CocoaPods React dependency would then depend on React-ART, and React-RCTSettings etc.
+
+## Motivation
+
+For teams adding React Native to existing apps CocoaPods support let's RN be treated as a dependency like all of the other ones. CocoaPods is built with a lot of escape valves for consumers to fix issues at runtime (you see people passing around `post_install` [fixes](https://github.com/facebook/react-native/issues/20182#issuecomment-409131862) for example, and I built a [CocoaPods plugin](https://github.com/orta/cocoapods-fix-react-native) to try consolidate them all)
```
eliperkins: s/support let's/support, this lets

proposals/0003-cocoapods-support-improvements.md
eliperkins: That is to say, we've had [bugs such as this](https://github.com/facebook/react-native/issues/19569) which were a result of an Xcode project being misconfigured. How will changing this CocoaPods structure allow for dependency changes like these to be addressed?

axe-fb:
cc @dehlen for the issue created at #15

proposals/0003-cocoapods-support-improvements.md
shergin: I believe man-made Xcode projects in the repo must die, they must be build artifacts of Buck packages. Nobody should configure builds in more than one place.

shergin:
I love the idea of decoupling dependencies and slicing existing huge CocoaPod file!

Do we have a strategy to address the fact that some deps of RN does not have Cocoapod version?

Let's consider Folly, for example. We are starting to use Folly more and more in the our new efforts and we have to have ability to depend on its the latest version available. Even we if can hack a Cocoapod for Folly (we actually tried and failed), we cannot expect that Folly contributors will fix the Cocoapod-based configuration every time they change something. So, I believe, we have to find a way to produce a Folly Cocoapod automatically from existing Buck (or cmake) configuration.

Same story with Yoga.

So, I would start the slicing from decoupling those.

proposals/0003-cocoapods-support-improvements.md
```diff
@@ -0,0 +1,117 @@
...
+
+# RFC0003: CocoaPods Support Improvements
+
+## Summary
+
+Simplify CocoaPods support in React Native, by reducing the differences between how CocoaPods maps React Native into the existing set of Xcode projects.
+
+## Basic example
+
+CocoaPods support in React Native is a tad tricky, because it’s trying to map a complex set of Xcode projects into a single dependency. The Podspec for React Native does a lot of work under the hood at ~300 LOC. This is, in part, because we use a CocoaPods abstraction called subspecs to consolidate out the React Native library into one single library.
```
shergin: I would also start from killing those hand-written projects. We have to figure out a way to build them automatically from Buck files which should be (they actually are) source of truth of the build configuration.

orta: I read up some of the buck docs - this *might* be possible by having buck output the compiler options into JSON and then having the pod specs capture that, I need to try prototype it to be sure though

cpojer:
@orta seems like people are in favor of this. What's the hold-up? Can you just go ahead and make this happen?

orta:
Yep, it can happen. 

It's mainly been blocked on us upgrading to the latest versions of React Native at Artsy (  https://github.com/artsy/emission/pull/1210 ) so that we can use/test it in-house. Which is on-going, and sporadically worked on by a few of us in our spare time. So, it might take some time for me to get around to it

cpojer:
I merged this RFC then (feel free to follow-up with small changes, given that you were still iterating on it a bit). I'll let you do this whenever you get around to do it.

PR #49, opened 2018-11-02
Move iOS and Android Specific Code to Their Own Packages

empyrical:
This is a proposal for a new structure the React Native project could take. I consider it to be part of the spirit of the slimmening, and could also lead to more components and apis being able to be moved out of core.

I have tried to describe as much as I could, but if there is anything else that should be described in this I would be happy to add to it!

I am happy to champion this effort and this would probably be the biggest open source contribution I've done yet, so it will be quite the learning experience.

### DIFF
--- a/proposals/0000-Move-Platforms-to-Own-Packages.md
+++ b/proposals/0000-Move-Platforms-to-Own-Packages.md
@@ -1,0 +1,273 @@
+---
+title: Move iOS and Android Specific Code to Their Own Packages
+author:
+- Ian Meikle
+date: today
+---
+
+# RFC0000: Move iOS and Android Specific Code to Their Own Packages
+
+## Summary
+
+Re-organize the React Native repo as a Yarn workspace monorepo, and move platform specific JS and native code for iOS and Android into their own packages.
+
+## Basic example
+
+Common APIs and components would remain in `react-native`:
+
+```js
+import {AppRegistry, Alert, View} from 'react-native';
+```
+
+Platform-specific components and APIs would be imported from the platform's module:
+
+```js
+// iOS
+import {AlertIOS} from 'react-native-ios';
+
+// Android
+import {PermissionsAndroid} from 'react-native-android';
+```
+
+## Motivation
+
+React Native is for more than just iOS and Android - there are renderers for Windows, macOS, and even the DOM and even more on their way.
+
+This is analogous to when React [split into React and ReactDOM](https://reactjs.org/blog/2015/09/10/react-v0.14-rc1.html#two-packages-react-and-react-dom), and (in my opinion at least!) the separation of these modules was a benefit to the renderers of other platforms.
+
+Splitting iOS and Android out into their own packages would make them functionally like out-of-tree platforms. The process of doing this split should hopefully strengthen support for out-of-tree platforms as a result.
+
+The process of re-organizing these modules will, I hope, also result in more components being found that can be split out into their own repo and package as part of the slimmening.
+
+Platform-specific APIs could drop the platform suffix from their name. `PermissionsAndroid` could just be `Permissions` for example.
+
+React Native being a Yarn workspace might make it easier to make other modules that should be developed and released alongside the main project. It would be easier to make a more advanced RNTester app in-tree with its own set of dependencies, for example.
+
+The IOS-only and Android-only files and folders at the top level of the repo could get organized into their respective packages.
+
+## Detailed design
+
+### Module exports:
+
+**Components**
+
+ * `react-native`
+   * **Components**
+   * `AccessibilityInfo`
+   * `ActivityIndicator`
+   * `ART`
+   * `Button`
+   * `CheckBox` (I am unsure about this one. Android-only in main tree, but generic enough other platforms could implement it. Possible slimmening candidate?)
+   * `FlatList`
+   * `Image`
+   * `ImageBackground`
+   * `ImageEditor`
+   * `ImageStore`
+   * `KeyboardAvoidingView`
+   * `ListView` (DEPRECATED. may be removed by the time of the split)
+   * `Modal`
+   * `Picker`
+   * `SafeAreaView`
+   * `ScrollView`
+   * `SectionList`
+   * `Slider`
+   * `Switch`
+   * `RefreshControl`
+   * `StatusBar`
+   * `SwipeableFlatList`
+   * `SwipeableListView` (DEPRECATED)
+   * `Text`
+   * `TextInput`
+   * `Touchable`
+   * `TouchableHighlight`
+   * `TouchableNativeFeedback`
+   * `TouchableOpacity`
+   * `TouchableWithoutFeedback`
+   * `View`
+   * `VirtualizedList`
+   * `WebView` (DEPRECATED. [Moved to its own repo](https://github.com/react-native-community/react-native-webview))
+   * **APIs**
+   * `Alert`
+   * `Animated`
+   * `AppRegistry`
+   * `AppState`
+   * `AsyncStorage`
+   * `BackHandler`
+   * `CameraRoll`
+   * `Clipboard`
+   * `DeviceInfo`
+   * `Dimensions`
+   * `Easing`
+   * `findNodeHandle`
+   * `I18nManager`
+   * `InteractionManager`
+   * `Keyboard`
+   * `LayoutAnimation`
+   * `Linking`
+   * `NativeEventEmitter`
+   * `NetInfo`
+   * `PanResponder`
+   * `PixelRatio`
+   * `Settings`
+   * `Share`
+   * `StyleSheet`
+   * `Systrace`
+   * `TVEventHandler`
+   * `UIManager`
+   * `unstable_batchedUpdates`
+   * `Vibration`
+   * `YellowBox`
+   * **Plugins**
+   * `DeviceEventEmitter`
+   * `NativeAppEventEmitter`
+   * `NativeModules`
+   * `Platform`
+   * `processColor`
+   * `requireNativeComponent`
+   * `takeSnapshot`
+   * **Deprecated PropTypes**
+   * `ColorPropType`
+   * `EdgeInsetsPropType`
+   * `PointPropType`
+   * `ViewPropTypes`
+
+ * `react-native-ios`
+   * **Components**
+   * `DatePickerIOS`
+   * `InputAccessoryView`
+   * `MaskedViewIOS`
+   * `PickerIOS`
+   * `ProgressViewIOS`
+   * `SegmentedControlIOS`
+   * `SnapshotViewIOS`
+   * `TabBarIOS`
+   * **APIs**
+   * `ActionSheetIOS`
+   * `AlertIOS`
+   * `ImagePickerIOS`
+   * `PushNotificationIOS`
+   * `StatusBarIOS`
+   * `VibrationIOS`
+
+ * `react-native-android`
+   * **Components**
+   * `DrawerLayoutAndroid`
+   * `ProgressBarAndroid`
+   * `ToastAndroid`
+   * `ToolbarAndroid`
+   * `ViewPagerAndroid`
+   * **APIs**
+   * `DatePickerAndroid`
+   * `PermissionsAndroid`
+   * `TimePickerAndroid`
+
+
+### Hypothetical project overview (non-exhaustive)
+
+This is the *very basic* overview of what the GitHub repo could look like:
+
+```
+-| project root
+--📁 {.circleci, .github, bots}/
+--📁 flow/
+--📁 flow-github/
+--📁 packages/
+---📁 react-native/
+----📁 Libraries/ (All cross-platform JS)
+----📁 ReactCommon/
+----📁 scripts/
+-----📄 launchPackager.bat
+-----📄 launchPackager.command
+-----📄 packager.sh
+----📁 third-party/
+----📁 tools/ (contains bzl build-defs)
+---📁 react-native-ios/
+----📁 Libraries/
+----📁 React/
+----📁 scripts/
+-----📄 ios-configure-glog.sh
+-----📄 ios-install-third-party.sh
+-----📄 react-native-xcode.sh
+----📁 third-party-podspecs/
+---📁 react-native-android/
+----📁 gradle/
+----📁 keystores/
+----📁 Libraries/
+----📁 ReactAndroid/
+---📁 react-native-git-upgrade/
+--📁 RNTester/
+--📁 scripts/ (CI, sync, etc. scripts)
+--📄 package.json (for the Yarn workspace)
+--📄 {readme, contributing, CoC, other markdown docs}
+```
+
+I am not fully sure what the layout of the local-cli stuff will end up needing to be after this; with [this proposal](https://github.com/react-native-community/discussions-and-proposals/blob/master/proposals/0002-Extracted-Command-Line-Tools.md) the bulk of it will end up getting moved to another repo anyways.
+
+ReactIOS and ReactAndroid would get entries in their `package.json` like this:
+
+```json
+{
+  "rnpm": {
+    "haste": {
+      "providesModuleNodeModules": ["react-native-ios"],
+      "platforms": ["ios"]
+    }
+  }
+}
+```
+
+Platform-specific Flow prop types will probably need to stay bundled together in the main package. Ideally it could be set up so if you didn't have `react-native-ios` installed the iOS-only View props wouldn't show up when typechecking, for example. But I don't know if there's a way to do that with Flow yet. Would need [this issue](https://github.com/facebook/flow/issues/396) to be resolved, probably.
+
+The native build scripts (gradle, BUCK, xcode, etc.) will all need to be updated for the new code locations. Great care must be taken to not break the existing large ecosystem of native compnents on NPM!
+
+## Drawbacks
+
+ - A lot of old open pull requests will need to be rebased if the codebase is re-arranged
+ - The `react-native-ios` and `react-native-android` packages on NPM already have the following versions on them:
+   > - `react-native-ios`: `1.0.0`, `1.0.1`, `1.1.0`
+   > - `react-native-android`: `1.0.0`, `2.0.0`, `3.0.0`
+   
+   This will make React Native's future semver bump to `1.0.0` awkward, if not impossible. Unless it does what React did and jumps to version `60.0.0` or something like that. Or starts at version `4.0.0`, which seems kinda random. NPM also says that for various technical and security reasons, these versions can not be deleted and re-used. I've `npm deprecate`d them, which is the best I can do sadly.
+ - While community modules that want to cover multiple versions of RN could use the "bridge" versions of the `{ios,android}` modules to ease the transition there are packages that are not maintained that would get broken by a split like this. This is also an issue for Slimmened modules as well.
+ - Even with a codemod to ease the transition, I don't imagine that transitioning codebases over will be something that is trivial. Maybe more than 2 versions should be given as a grace period before everything moves over?
+ - Old documentation and tutorials will need to be updated.
+
+## Alternatives
+
+I believe that this is the best way forward for helping improve React Native's third party platform support, and also for helping improve the existing in-tree platforms by organizing platform-specific things into their own packages. If it's not, hopefully this will spawn discussions on better alternatives.
+
+## Adoption strategy
+
+There could be "bridge" versions of the `react-native-{ios,android}` modules for pre-split versions of React Native that simply import platform specific stuff from `react-native`. This would make it easier for people to make React Native addons that cover wider ranges of RN versions
+
+For a release or so, `react-native-{ios,android}` should probably be thin modules that just re-import stuff from `react-native` before the code actually gets moved over, to help people transition.
+
+For example, the import for `DatePickerIOS` in `react-native` would look like:
+
+```js
+  get DatePickerIOS() {
+    warnMoved('react-native-ios', 'DatePickerIOS');
+    return require('DatePickerIOS');
+  },
+```
+
+(where `warnMoved` is a hypothetical function warning once that an api is moved to another module)
+
+And `react-native-ios` would use Haste to import the module from `react-native`:
+
+```js
+  get DatePickerIOS() {
+    return require('DatePickerIOS');
+  },
+```
+
+Code-mods could be written to help with the transition. React and ReactDOM could be looked to as an example on how to do a transition like this as painlessly as possible.
+
+## How we teach this
+
+A documentation page layout out where all these components are getting moved around to should be made. This page could also document where Slimmen'd modules got moved to.
+
+## Unresolved questions
+
+ * After the platform-specific APIs are moved to the platform module, should the platform suffix be removed from the named export? (The filename may need to keep the suffix for now, to avoid Haste collisions)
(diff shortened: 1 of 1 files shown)


proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Basic example
+
+Common APIs and components would remain in `react-native`:
+
+```js
+import {AppRegistry, Alert, View} from 'react-native';
+```
+
+Platform-specific components and APIs would be imported from the platform's module:
```
TheSavior: What about things that should be cross platform but just haven’t been implemented on the other platform yet? Do they start in a platform specific package and move into the core package later?

Or is it more based on the particular api/component in which case there would still be some things in core that don’t work on both platforms?

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+
+This is analogous to when React [split into React and ReactDOM](https://reactjs.org/blog/2015/09/10/react-v0.14-rc1.html#two-packages-react-and-react-dom), and (in my opinion at least!) the separation of these modules was a benefit to the renderers of other platforms.
+
+Splitting iOS and Android out into their own packages would make them functionally like out-of-tree platforms. The process of doing this split should hopefully strengthen support for out-of-tree platforms as a result.
+
+The process of re-organizing these modules will, I hope, also result in more components being found that can be split out into their own repo and package as part of the slimmening.
+
+Platform-specific APIs could drop the platform suffix from their name. `PermissionsAndroid` could just be `Permissions` for example.
+
+React Native being a Yarn workspace might make it easier to make other modules that should be developed and released alongside the main project. It would be easier to make a more advanced RNTester app in-tree with its own set of dependencies, for example.
```
TheSavior: RNTester being its own package in-tree is unrelated and could happen anyways right? Perhaps that should be part of the RNTester proposal.

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+   * **Deprecated PropTypes**
+   * `ColorPropType`
+   * `EdgeInsetsPropType`
+   * `PointPropType`
+   * `ViewPropTypes`
+
+ * `react-native-ios`
+   * **Components**
+   * `DatePickerIOS`
+   * `InputAccessoryView`
```
TheSavior: This is an example of something I could imagine being standard across many platforms. Also, should probably be pulled out with slimmening regardless.

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+  "rnpm": {
+    "haste": {
+      "providesModuleNodeModules": ["react-native-ios"],
+      "platforms": ["ios"]
+    }
+  }
+}
+```
+
+Platform-specific Flow prop types will probably need to stay bundled together in the main package. Ideally it could be set up so if you didn't have `react-native-ios` installed the iOS-only View props wouldn't show up when typechecking, for example. But I don't know if there's a way to do that with Flow yet. Would need [this issue](https://github.com/facebook/flow/issues/396) to be resolved, probably.
```
TheSavior: How do other platforms like windows extend the props of components to provide platform specific props?

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Alternatives
+
+I believe that this is the best way forward for helping improve React Native's third party platform support, and also for helping improve the existing in-tree platforms by organizing platform-specific things into their own packages. If it's not, hopefully this will spawn discussions on better alternatives.
+
+## Adoption strategy
+
+There could be "bridge" versions of the `react-native-{ios,android}` modules for pre-split versions of React Native that simply import platform specific stuff from `react-native`. This would make it easier for people to make React Native addons that cover wider ranges of RN versions
+
+For a release or so, `react-native-{ios,android}` should probably be thin modules that just re-import stuff from `react-native` before the code actually gets moved over, to help people transition.
```
TheSavior: Or the other way where React Native keeps exposing things from the native packages but the changes can be made in the native packages directly.

proposals/0000-Move-Platforms-to-Own-Packages.md
empyrical: I'm not sure if that could be done without re-arranging it into a monorepo first, Yarn requires the top level `package.json` of a repo to be private in a yarn workspace.

But to implement this idea, I was thinking I would first do a "smaller" pull request that *just* sets up a yarn workspace and nothing else, and that could get done sooner. Maybe the yarn workspace could be a separate proposal this one is dependant on?

proposals/0000-Move-Platforms-to-Own-Packages.md
empyrical: > What about things that should be cross platform but just haven’t been implemented on the other platform yet?

I was thinking that an `UnimplementedView` stub should be in the main package and an implemented JS file would be in the platform module.

Like in the case of `CheckBox`, a `UnimplementedView` stub `CheckBox.js` would be in the main package, and a `CheckBox.android.js` in `react-native-android`

proposals/0000-Move-Platforms-to-Own-Packages.md
empyrical: Will move this to the main package when I revise this 👍

proposals/0000-Move-Platforms-to-Own-Packages.md
empyrical: In the case of RNWindows they use the platform-overridable `prop-types` files:

https://github.com/Microsoft/react-native-windows/blob/master/Libraries/Components/View/PlatformViewPropTypes.windows.js
https://github.com/Microsoft/react-native-windows/blob/master/Libraries/Text/TextPropTypes.windows.js

They don't seem to use platform overridable flow props anywhere. But a platform-specific `X-Types.platform.js` would be awkward, because you wouldn't be able to see the flow types for `windows` when flow checking `*.ios.js` files for example.

proposals/0000-Move-Platforms-to-Own-Packages.md
TheSavior: Honestly I think this particular one should just get moved to a community rep. But that is a conversation for slimmening. Nothing to worry about for this proposal.

empyrical:
I think that the first thing that should be done is land an initial set up of a Yarn workspace. It would both reveal possible places internally where things might break, and would also make it so RNTester could sit in the workspace without being dependant on the progress/acceptance of this proposal. I will look into doing a PR of this.

empyrical:
Do you mean, like, moving all platform-specific components into userland instead of splitting them out into platform modules?

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+```js
+  get DatePickerIOS() {
+    warnMoved('react-native-ios', 'DatePickerIOS');
+    return require('DatePickerIOS');
+  },
+```
+
+(where `warnMoved` is a hypothetical function warning once that an api is moved to another module)
+
+And `react-native-ios` would use Haste to import the module from `react-native`:
```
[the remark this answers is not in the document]
empyrical: Yeah, that would definitely also make using other packagers (like parcel and webpack) less complex too. @gaearon removed haste from react a while ago and that could be looked to as an example of what could be done:

https://github.com/facebook/react/pull/11303

Platform-specific overrides, however, happen to be really good at doing the platform-specific overrides across multiple packages. A good replacement solution for this would need to be devised. Definitely a subject worthy of a proposal/discussion of its own probably!

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+The native build scripts (gradle, BUCK, xcode, etc.) will all need to be updated for the new code locations. Great care must be taken to not break the existing large ecosystem of native compnents on NPM!
+
+## Drawbacks
+
+ - A lot of old open pull requests will need to be rebased if the codebase is re-arranged
+ - The `react-native-ios` and `react-native-android` packages on NPM already have the following versions on them:
+   > - `react-native-ios`: `1.0.0`, `1.0.1`, `1.1.0`
+   > - `react-native-android`: `1.0.0`, `2.0.0`, `3.0.0`
+   
+   This will make React Native's future semver bump to `1.0.0` awkward, if not impossible. Unless it does what React did and jumps to version `60.0.0` or something like that. Or starts at version `4.0.0`, which seems kinda random. NPM also says that for various technical and security reasons, these versions can not be deleted and re-used. I've `npm deprecate`d them, which is the best I can do sadly.
```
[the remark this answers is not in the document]
empyrical: That's definitely an option! That idiom would make it awkward for `react-native-dom` however, since `react-dom` already exists.

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+   * `takeSnapshot`
+   * **Deprecated PropTypes**
+   * `ColorPropType`
+   * `EdgeInsetsPropType`
+   * `PointPropType`
+   * `ViewPropTypes`
+
+ * `react-native-ios`
+   * **Components**
+   * `DatePickerIOS`
```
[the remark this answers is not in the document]
empyrical: `DatePickerIOS` is a component, while `DatePickerAndroid` is an API that pops up a little modal. *Might* be possible, but would be a bit of work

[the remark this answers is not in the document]
empyrical: The IOS one is a component, while the Android one is an API that opens a dialog. Might be possible but would be tricky

proposals/0000-Move-Platforms-to-Own-Packages.md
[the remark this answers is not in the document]
necolas: I think `react-dom` as a package name makes sense for what it does, and wouldn't make sense as a name for a package in this context. We'd only want a single web implementation, so I don't think we need to worry about mapping every OSS RN variant to an equivalent package in RN.

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+----📁 Libraries/
+----📁 ReactAndroid/
+---📁 react-native-git-upgrade/
+--📁 RNTester/
+--📁 scripts/ (CI, sync, etc. scripts)
+--📄 package.json (for the Yarn workspace)
+--📄 {readme, contributing, CoC, other markdown docs}
+```
+
+I am not fully sure what the layout of the local-cli stuff will end up needing to be after this; with [this proposal](https://github.com/react-native-community/discussions-and-proposals/blob/master/proposals/0002-Extracted-Command-Line-Tools.md) the bulk of it will end up getting moved to another repo anyways.
```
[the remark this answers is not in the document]
empyrical: > OSS RN packages being published without being compiled

IMO I kinda like them shipping un-compiled. Especially with JSI allowing for different JS engines to be used with React Native, it would be useful for it to be easy to use different babel configs for different engines (imo). Biggest thing would be opting in to using native async/await where available for a cleaner error stack trace

Would be worth a discussion of its own for sure!

proposals/0000-Move-Platforms-to-Own-Packages.md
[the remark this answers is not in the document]
empyrical: Def. worth making another issue for, since your idea of having information about out-of-tree platforms in the main repo could probably be done independently of this proposal.

> We'd only want a single web implementation

RNWeb and RNDom are different enough projects I think there's room for the both of them. But I really don't know about dropping the `native` from the name, because a "React renderer" and a "React Native renderer/platform" are very different things.

proposals/0000-Move-Platforms-to-Own-Packages.md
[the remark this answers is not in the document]
necolas: There's a difference between not compiling parts of the code that are effectively js standards but not available in older environments, and publishing code with early stage proposals

kevinvangelder:
Not sure if this should be part of this proposal or not, but other people are discussing it, so I'll throw in my two cents. I think we should avoid ever naming modules with the iOS or Android suffix. Even if the other platform doesn't have the necessary APIs available to complete the module's objective it might in the future, causing a rename when it's implemented, or there may be less-common platforms, such as Windows, that do have the necessary APIs that have to implement the awkward module naming for purposes of supporting cross-platform use of the module. I gather the original purpose of doing this was to make it clear that the module only supported particular platforms even for people too lazy to read the docs, but I don't think the possibility of third-party platforms was considered at the time.

proposals/0000-Move-Platforms-to-Own-Packages.md
acoates-ms: I think the core should be slim enough that its pretty much required to have it all implemented.  So for instance, checkbox wouldn't be part of core.

empyrical:
I highly agree that the platform suffix should be dropped from the names! 👍 

However in some cases a platform will need a completely different implementation of a component/API and can't be generic. In cases like that it could just go in the platform module, rather than in the main module. `DatePicker...` is a good example of this. It's a component on IOS, but an API that summons a dialog on Android.

kevinvangelder:
Right, but hiding implementation details is exactly what React Native is good at, so as long as the behavior isn't wildly unexpected.

empyrical:
If there are no objections, I will be revising the proposal to drop the platform suffix from the moved APIs

proposals/0000-Move-Platforms-to-Own-Packages.md
empyrical: Agreed, would be good to keep core to a minimum "This is what you need to implement to be considered a full React Native implementation". Other stuff could be implemented by the platform implementers by submitting pull requests to the Slimmen'd repos (like what will probably happen with Windows' WebView implementation)

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,273 @@
...
+
+# RFC0000: Move iOS and Android Specific Code to Their Own Packages
+
+## Summary
+
+Re-organize the React Native repo as a Yarn workspace monorepo, and move platform specific JS and native code for iOS and Android into their own packages.
+
+## Basic example
+
+Common APIs and components would remain in `react-native`:
```
grabbou: How Windows and other platforms would provide their implementations of core components and APIs? Since imports are from `react-native`, not from platform-specific module, I guess that would be tricky and require certain platforms to export a minimal subset of APIs to be considered a valid platform for React Native?

proposals/0000-Move-Platforms-to-Own-Packages.md
grabbou: What about a module that has 4 methods out of 5 that work on both platforms and wants to provide one, additional, e.g. `Alert.promptIOS`? Would the developer need to make an additional `Alert` module under `react-native-ios` package? Wouldn't that cause troubles for developers to decide what file to use? Should `Alert` from `react-native-ios` "extend" `Alert` from `react-native` to keep its methods? Just general questions

proposals/0000-Move-Platforms-to-Own-Packages.md
empyrical: Platforms can have an entry like this in their `package.json` which adds their platform to the haste search path:

```json
{
  "rnpm": {
    "haste": {
      "providesModuleNodeModules": ["react-native-example"],
      "platforms": ["example"]
    }
  }
}
```

And if they wanted to override `ExampleComponent` in `react-native`, `react-native-example` would have `ExampleComponent.example.js`

This solution depends on Haste, so if RN wants to drop Haste in the future a solution for platform overrides will need to be worked out. 

> I guess that would be tricky and require certain platforms to export a minimal subset of APIs to be considered a valid platform for React Native

Pretty much, or have a whole bunch of stub JS `UnimplementedView` files to address the Haste missing files errors

proposals/0000-Move-Platforms-to-Own-Packages.md
empyrical: I do think some kinds of revamp of `Alert` is very much needed. There's this `if` statement in `Alert` that basically requires a platform dev to make a platform fork of `Alert`. Example from RNWindows:

```js
    if (Platform.OS === 'ios') {
      if (typeof type !== 'undefined') {
        console.warn('Alert.alert() with a 5th "type" parameter is deprecated and will be removed. Use AlertIOS.prompt() instead.');
        AlertIOS.alert(title, message, buttons, type);
        return;
      }
      AlertIOS.alert(title, message, buttons);
    } else if (Platform.OS === 'android') {
      AlertAndroid.alert(title, message, buttons, options);
    } else if (Platform.OS === 'windows') {
      AlertWindows.alert(title, message, buttons, options);
    }
```

And because it depends on requiring `AlertIOS`, it looks like some kind of modification will need to be made to this file in the split.

kevinvangelder: Doing a suffix on the method is still bad because just because Android doesn't support it doesn't mean that Windows/MacOS/Ubuntu/other third party platforms won't. Rather than baking it into the method name it should just be documented which platforms it supports.

proposals/0000-Move-Platforms-to-Own-Packages.md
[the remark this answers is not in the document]
victoriafrench: what about going babel’s route and going namespaced.

@react-native/core
@react-native/ios
@react-native/android
@react-native/windows
@react-native/tvos

it would allow for semver 1.0.0 and identify official packages ie react-native-web would immediately be identifiable as an external project.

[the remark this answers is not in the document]
victoriafrench: @empyrical I ran a namespace search and it appears the orgid [react-native](https://www.npmjs.com/~react-native) is taken (0 public releases may be private only) but orgs [react](https://www.npmjs.com/~react) and [reactnative](https://www.npmjs.com/~reactnative) are not. Maybe worth snagging them now even if they will not be used.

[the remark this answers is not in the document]
empyrical: no worries! FWIW, facebook DOES own `@react` now. I could see if the owner of `@reactnative` would be open to transferring it, but minus the dash `-` is a bit weird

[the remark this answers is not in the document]
victoriafrench: I couldn't see who owned @react because the profile was empty. I reached out to npm asking to be pointed to who has both @react and @reactnative earlier, pointing them here.

If @react is now under facebooks control, if they added both the react and react native community to the org... this would be a great naming convention if everyone got on board:

@react/core
@react/native-core
@react/prop-types
@react/dom
@react/ios
@react/android
@react/web
@react/ etc etc etc

Very straightforward.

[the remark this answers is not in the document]
empyrical: I've learned that RN and React might be converging a bit more, so that (and @necolas' suggestion to use `react-ios` as well) would make sense 👍 

I've updated the proposal

proposals/0000-Move-Platforms-to-Own-Packages.md
```diff
@@ -0,0 +1,277 @@
...
+
+Splitting iOS and Android out into their own packages would make them functionally like out-of-tree platforms. The process of doing this split should hopefully strengthen support for out-of-tree platforms as a result.
+
+The process of re-organizing these modules will, I hope, also result in more components being found that can be split out into their own repo and package as part of the slimmening.
+
+Platform-specific APIs could drop the platform suffix from their name. `PermissionsAndroid` could just be `Permissions` for example.
+
+React Native being a Yarn workspace might make it easier to make other modules that should be developed and released alongside the main project. It would be easier to make a more advanced RNTester app in-tree with its own set of dependencies, for example.
+
+(**UPDATE**: I am working on a pull request to set up a monorepo, so other proposals could benefit from it regardless of this proposal's progress)
```
cpojer: Just fyi: This can probably not be done on GitHub and will have to be done at FB in one go.

grabbou:
Update: We have split CLI into multiple packages (android, ios and core) and it's been the best decision in a while. 

I was wondering if there's any outcomes from this PR, such as we're unlikely to implement this architecture or eventually, that's the direction we're going?

TheSavior:
I think the people on our team at FB want these pieces separated as well. However, this is major refactoring work that can only be accomplished internally, so it’s probably a ways out. Probably at least a year until anything could actually happen here unfortunately. We have lots of other projects we have to focus on first. :-)

kevinvangelder:
Even if it can't happen for a while, it would be good to solidify the plan.

larakitb:
> I think the people on our team at FB want these pieces separated as well. However, this is major refactoring work that can only be accomplished internally, so it’s probably a ways out. Probably at least a year until anything could actually happen here unfortunately. We have lots of other projects we have to focus on first. :-)

Is there any updates from Meta team around this ?

cortinico:
> Is there any updates from Meta team around this ? Thanks!

What @TheSavior mentioned above still stands:

> I think the people on our team at FB want these pieces separated as well. However, this is major refactoring work that can only be accomplished internally, so it’s probably a ways out. Probably at least a year until anything could actually happen here unfortunately. We have lots of other projects we have to focus on first. :-)

I don't think we have major updates to share on this side. I believe @rickhanlonii had some opinion on this project and maybe has more to share.

Sharing my personal point of view here: I think that reviving this proposal and trying to land it in some form would be highly beneficial. I feel that the current setup of the repo is slowing us down (i.e. it's complicated to export code to separate packages) + it's hard to understand for newcomers (i.e. why iOS code is inside `React` and Android code inside `ReactAndroid`?).

As this RFC looks like a big refactoring, I share all the concerns that were already presented: it should probably be done internally, or at least will require some guidance and coordination from Meta.

To make it a bit more actionable, I believe we could split the whole RFC in three big milestones:

1. **Re-organize the project into a proper Yarn workspace** - Essentially, having `react-native` as a top level package and having `repo-config/package.json` as our workspace is really suboptimal. It introduces a lot of complexity and makes it hard to extract anything inside the `package/` folder we currently have.  

2. **Align versioning of `react-native` and the other packages we currently have, and let the CI publish them** - Currently the versions of `react-native` we have on `main` is updated from `1000.0.0` to the needed one whenever we publish a nightly/release/release-candidate. That's not true for the other NPM packages inside `packages/` folder. Those packages are following arbitrary version schemas and are published on a "per-need" basis. This means that our nightly versions of RN are not actually using all the code we have on `main`. Only the `react-native` package is doing so, other dependencies (i.e. `react-native-codegen`) are not updated and we might have code on `main` for them, which is not actually published. I think that we should align the versioning of all the packages to the RN one. Moreover, we should potentially move to having the CI publish all the packages, not only `react-native`.

3. **Modularize `react-native`**. I.e. Split the Android/iOS code into separate packages (`react-native-ios`, `react-native-android` and more).

Point 3. seems to introduce a lot of complexity due to implications on how we treat APIs that are platform specific, etc. However, I think that 1. and 2. are potentially doable and would be a first step towards unblocking a proper modularization.

Curious to hear other's opinion here 👍

vksgautam1:
any news ?

PR #54, opened 2018-11-15
Proposal to add additional accessibility roles and states.

marcmulcahy:
This is the first of several proposals to improve the cross-platform accessibility of React Native. This proposal describes the fairly simple addition of some missing useful accessibility roles and states which are present in many platforms.

### DIFF
--- a/proposals/0000-a11y-roles-states-improvements.md
+++ b/proposals/0000-a11y-roles-states-improvements.md
@@ -2,7 +2,7 @@
 title: Improvements to accessibilityRole and accessibilityStates
 author:
 - Marc Mulcahy
-date: 2018-10-31
+date: 2019-02-06
 ---
 
 ## Summary
@@ -41,6 +41,7 @@ This proposal adds the following roles:
 - radiogroup
 - scrollbar
 - spinbutton
+- switch
 - tab
 - tablist
 - timer
@@ -54,6 +55,10 @@ This proposal adds the following states:
 - off
 - checked
 - unchecked
+- busy
+- expanded
+- collapsed
+- hasPopup
 
 ### Fallback Mechanism for Roles
 
@@ -61,7 +66,7 @@ There are several roles which don't have native analogs on a given platform. For
 
 #### Android
 
-The fallback mechanism is already in place on Android. For roles which don't have standard Android view analogs, the class name of the AccessibilityNodeInfo is set to an arbitrary value, and the roleDescription extra is set to a localized textual description of the role.
+The fallback mechanism is already in place on Android. For roles which don't have standard Android view analogs, the class name of the AccessibilityNodeInfo is set to the class name of the closest native Android view type, and the roleDescription extra is set to a localized textual description of the role.
 
 ### Fire OS
 
@@ -74,13 +79,21 @@ The role string should not be localized-- it should be one of the constants defi
 
 #### iOS
 
-For roles which don't have a corresponding iOS accessibilityTrait, we recommend appending a localized version of the role description to the accessibility label.
+For roles which don't have a corresponding iOS accessibilityTrait, we recommend appending a localized version of the role description to the accessibilityValue of the component.
 
 ### Additional States Implementation Notes
 
-Android has slightly more robust state support than does iOS. Both the new checked and on states will be exposed directly by setting the isCheckable and isChecked members of the corresponding AccessibilityNodeInfo.
+Android has slightly more robust state support than does iOS. The following states can be represented with Android'd standard accessibility API:
 
-on iOS, the solution is less clear. Unless a better solution is found, we propose to add a textual description of the state to the accessibilityLabel.
+- checked
+- unchecked
+- expanded
+- collapsed
+- hasPopup
+
+We will append a localized description of other states to the component's content description.
+
+on iOS, a localized description of states which do not have corresponding accessibilityTraits will be appended to the component's accessibilityValue.
 
 ## Drawbacks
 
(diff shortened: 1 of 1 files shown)


proposals/0000-a11y-roles-states-improvements.md
```diff
@@ -0,0 +1,106 @@
...
+- radiobutton
+- radiogroup
+- scrollbar
+- spinbutton
+- tab
+- tablist
+- timer
+- toolbar
+
+### New States
```
necolas: Web has a lot more states that we'll need to bring over, including `busy`, `expanded`, `hasPopup` etc. I don't know how `on` and `off` map to web ARIA states. Any thoughts?

proposals/0000-a11y-roles-states-improvements.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Motivation
+
+When developers build components in JavaScript, they need the ability to express the role and current state(s) of the component to assistive technologies running on the platform.
+
+## Detailed design
+
+### New Roles
+
+This proposal adds the following roles:
```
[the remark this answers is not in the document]
marcmulcahy: Agree. I wonder if a good next step might be to create a master list of roles and states, based on ARIA, and then map those to native roles/states where appropriate.

proposals/0000-a11y-roles-states-improvements.md
marcmulcahy: Not sure either ATM. As I mentioned above, maybe we start with ARIA for roles and states, and then add additional states as needed.

proposals/0000-a11y-roles-states-improvements.md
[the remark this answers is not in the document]
jessebeach: @necolas here's a complete list of ARIA roles and their state/properties (as determined by inheritance) in JSON format: https://gist.github.com/jessebeach/3cf49ecac439ac6b5fd42935b5c43278

I pulled the information from my project, aria-query: https://github.com/A11yance/aria-query

proposals/0000-a11y-roles-states-improvements.md
```diff
@@ -0,0 +1,106 @@
...
+
+Some brilliant work was begun to add cross platform support for accessibility roles and states to React Native. We propose to add some additional roles and states, as well as extend the implementation on Android and Fire OS.
+
+## Basic example
+
+With the current set of roles, it is not possible to describe a checkbox. These changes would enable this scenario, among other things:
+
+    <View
+      accessibilityRole="checkbox"
+      accessibilityStates={['unchecked']}>
```
jessebeach: Would `'unchecked'` be drawn from an enumerable?

proposals/0000-a11y-roles-states-improvements.md
```diff
@@ -0,0 +1,106 @@
...
+- scrollbar
+- spinbutton
+- tab
+- tablist
+- timer
+- toolbar
+
+### New States
+
+This proposal adds the following states:
```
jessebeach: `disabled`

proposals/0000-a11y-roles-states-improvements.md
marcmulcahy: Yes, these are the the only states we've identified for now. ARIA supports a bunch of states, and it's not always clear how they're supposed to be used. Let's start small, and then add states and support as we find compelling need.

As for on and off, they make sense for things like switches, where the closest analog is probably checked/unchecked, but it doesn't make sense to present something like a light as "checked". This prompts me to think we probably need a "switch" role as well.

As for how to map on/off to web, probably checked/unchecked is the closest...

proposals/0000-a11y-roles-states-improvements.md
marcmulcahy: Looks like disabled is already supported?

proposals/0000-a11y-roles-states-improvements.md
marcmulcahy: I'd prefer not to go this route, as it makes implementation a lot more complex, in that each state has to be an object with variable type. Any reason to add such complexity?

proposals/0000-a11y-roles-states-improvements.md
```diff
@@ -0,0 +1,106 @@
...
+
+We should document the roles as part of the official accessibility API documentation. Developers can then start using them if they choose.
+
+## How we teach this
+
+The attempt with this proposal is to bring React Native accessibility support one step closer to generally accepted accessibility idioms and practices as specified by standards such as ARIA. As such, we believe this will make it easier for developers familiar with accessibility on the web or other platforms to relate React Native's accessibility support to what they already know.
+
+## Unresolved questions
+
+- Roles imply supported states and behaviors. For example, checkboxes imply the checked state, and the ability to be toggled. Radio groups typically only allow one child to be selected, etc. When a component specifies its role, should this programmatically require the developer to implement certain behaviors? Alternatively, what mechanisms exist to present warnings during compile time for missing behaviors?
```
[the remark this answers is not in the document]
marcmulcahy: Yes, Android's support for roles and states is a bit more robust than is iOS's, but for now, many of these new roles will simply resolve to adding info to the presentation of the object. Our forthcoming proposal and work on improving accessibility actions support will allow developers to ore accessibly support complex touch-screen interactions. Don't think ARIA has a solution for accessible actions as of yet....

proposals/0000-a11y-roles-states-improvements.md
jessebeach: @marcmulcahy as long as the strings are arbitrary. If not, if field values are drawn from a specific set, then without an enum, the functionality is simply based on convention.

proposals/0000-a11y-roles-states-improvements.md
jessebeach: @necolas `hasPopup` is a property in the sense that it designates the possibility of a popup; it is not meant to indicate that a popup is visible or not.

proposals/0000-a11y-roles-states-improvements.md
jessebeach: @marcmulcahy I think I'm imposing a implementation detail on the spec. I retract these comments.

PR #55, opened 2018-11-15
Proposal to add accessibility actions.

marcmulcahy:
Many cross-platform components require interactions beyond simple touches. A simple example of such a component is a slider, which requires the ability to adjust the value by sliding a finger. Other examples include swipe actions (swipe right to delete an email message for example).

To allow disabled users to access these functions of a component, accessibility actions support is required.

### DIFF
--- a/proposals/0000-a11y-actions.md
+++ b/proposals/0000-a11y-actions.md
@@ -1,0 +1,84 @@
+---title: Accessibility Actions Support
+author:
+- Marc Mulcahy
+date: 2018-10-31
+---
+
+# RFC0000: Accessibility Actions Support
+
+## Summary
+
+Custom components often require interactions beyond simple touches. Programatic access to these interactions must be provided to allow disabled users to perform them.
+
+## Basic example
+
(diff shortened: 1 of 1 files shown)


proposals/0000-a11y-actions.md
```diff
@@ -0,0 +1,84 @@
...
+* adjustup: Adjusts component up
+* adjustdown: adjust component down
+* expand: Expands the component
+* collapse: Collapses the component
+* copy: copies content in the component to the clipboard
+* cut: cuts content in the component to the clipboard
+* paste: pastes clipboard contents into this component
+* setvalue: sets the value of this component (typically used to adjust sliders, etc.)
+
+### Implementation Notes
```
necolas: What's the proposed story for web? I don't think there are any equivalent accessibility APIs

marcmulcahy: I don't think so either. But I think we'll need action support for native, especially given the behavior on Android 8.0 and above, where the framework no longer simulates a touch event to the center of a component's bounding box, but instead generates a click accessibility action. Most standard Android views handle this, but some React native views like InputText don't. Probably also applies to some handling in iOS, where clicking on the center of a bounding box doesn't make sense, like links which span multiple lines.

jessebeach: @necolas I can't come up with a scenario where we'd need the equivalent accessibility API on web, since we don't have a source of actions from an AX service like Native platforms have.

PR #56, opened 2018-11-15
Proposal for various cross platform a11y improvements.

marcmulcahy:
This proposal describes various cross platform improvements supported by most platforms.

### DIFF
--- a/proposals/0000-a11y-cross-platform.md
+++ b/proposals/0000-a11y-cross-platform.md
@@ -1,0 +1,133 @@
+---
+title: Cross-platform Accessibility Improvements
+author:
+- Marc Mulcahy
+date: 2018-10-31
+---
+
+# RFC0000: Cross-platform Accessibility Improvements
+
+## Summary
+
+Several accessibility-related properties are only available on one platform (either iOS or Android, but not both). This proposal attempts to make as many properties available on both Android and iOS as possible.
+
+Properties only available on iOS include:
+
+* accessibilityViewIsModal
+* accessibilityElementsHidden
+
+Properties only available on Android include:
+
+* accessibilityLiveRegion
+
+Several properties are missing from both platforms:
+
+* accessibilityCompoundComponent: replace accessible-- indicates to assistive technologies that all the components inside this one should be treated as a single component.
+* accessibilityRelationships: describes how this component is related to others
+* accessibilityRange: expose information about range-based components such as sliders
+
+## Motivation
+
+All accessibility properties should be available on all supported platforms. This reduces the burden on developers to understand the nuances of accessibility support on various platforms. It also allows JavaScript components to expose rich accessibility information, rather than relying on the accessibility support in native components.
+
+## Detailed design
+
+### accessibilityViewIsModal
+
+#### Android/Fire OS
+
+When a component marks itself as modal, code in the view manager will need to traverse all other components in the view tree and mark them as importantForAccessibility=no. A window content changed event with content change types set to subtree will then need to be fired on the root window to ensure that assistive technologies receive the tree updates.
+
+It may also be necessary to install a view tree observer once the modal component is active to prevent background components from appearing. We'll need to do some real-world testing to understand how realistic this scenario might be, and whether it's worth the performance cost of managing such an observer.
+
+#### iOS
+
+This property is currently supported on iOS, so no change to iOS support is needed.
+
+### accessibilityElementsHidden
+
+#### Android/Fire OS
+
+Supporting this property in Android may be as simple as marking the component as importantForAccessibility=noHideDescendants
+
+If this doesn't match the iOS behavior, we may need to build an accessibility delegate to hide the children. This will likely be tricky, since the children will still be generating events.
+
+#### iOS
+
+This property is already supported on iOS, so no iOS changes are required.
+
+### accessibilityLiveRegion
+
+This allows components to tell assistive technologies how to react to changes in their content.
+
+#### Android/Fire OS
+
+This property is already supported on Android, so no changes in Android support are required.
+
+#### iOS
+
+We believe this functionality can be implemented using accessibility notifications on iOS. For assertive live regions, the updated text of the component can be sent immediately to iOS as an announcement notification. For polite live regions, code can wait for the AnnouncementDidFinish notification before posting the updated text as an announcement notification.
+
+### accessibilityCompoundComponent
+
+We think the current implementation, which is called "accessible", should be renamed to accessibilityCompoundComponent, to more accurately represent what it does. The functionality should remain the same.
+
+### accessibilityRelationships
+
+This property describes how this component is related to others. Relationship types include:
+
+* labelFor: This component labels the specified component
+* labeledBy: this component is labeled by the specified component
+* describedBy: this component is described by the specified components.
+
+When referring to other components, we may need to add an additional property to view-- namely accessibilityId. We need to find the best forward-looking approach for referencing other components.
+
+#### Android/Fire OS
+
+For labelFor and labeledBy, use the standard Android view and AccessibilityNodeInfo methods.
+
+For describedBy, which is only supported on Fire OS, put the appropriate key/values in the extras bundle.
+
+#### iOS
+
+When associating labels with components, include the label text in the target component's accessibilityLabel.
+
+Don't support describedBy on iOS for now. In future, we could explore adding the accessibilityLabels of the describedBy targets to the accessibilityHint of the component.
+
+### accessibilityRange
+
+This property allows components (such as sliders) to expose range-based information. The value of the property is an object containing the following values:
+
+* min: the minimum value of the range
+* max: the maximum value of the range
+* current: the current value of the component (within the range specified by min and max)
+
+#### Android/Fire OS
+
+This maps directly to Android's RangeInfo interface.
+
+#### iOS
+
+iOS supports a property called accessibilityValue. This value is a simple string. accessibilityRange will be converted to a localized string and placed into accessibilityValue.
+
+## Drawbacks
+
+It will be challenging to ensure that the behavior of these properties is the same between iOS and Android. Building cross-platform properties will likely involve sacrificing a small amount of functionality on each target platform to ensure consistency.
+
+## Alternatives
+
+Components could continue to use the platform-specific properties. This has the drawback of requiring developers to have at least a cursory
+understanding of accessibility on both iOS and Android. And in the case of accessibilityViewIsModal, we expect this to be somewhat difficult to implement correctly on Android.
+
+## Adoption strategy
+
+We don't believe this to be a breaking change. Developers who currently use these properties on Android or iOS can continue to do so, and remove conditional use to allow functionality on other platforms.
+
+## How we teach this
+
+The same documentation that is used to describe the properties on either iOS or Android can be marked as cross-platform for use on either. For the accessibilityRelationships property, labeledBy and labelFor are common paradigms across accessibility support on multiple platforms including web.
+
+## Unresolved questions
+
+* Exactly what is the behavior for accessibilityElementsHidden on iOS? Does this hide the component itself from accessibility, or only its children?
+* What is the best mechanism for one component to refer to another for the purpose of specifying labels, targets, descriptive components, etc.?


proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+# RFC0000: Cross-platform Accessibility Improvements
+
+## Summary
+
+Several accessibility-related properties are only available on one platform (either iOS or Android, but not both). This proposal attempts to make as many properties available on both Android and iOS as possible.
+
+Properties only available on iOS include:
+
+* accessibilityViewIsModal
+* accessibilityElementsHidden
```
necolas: This is similar to Android `importantForAccessibility` prop. It would be good to consolidate these together and learn from how web handles these cases

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+Properties only available on Android include:
+
+* accessibilityLiveRegion
```
necolas: This is also available on web

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+Several properties are missing from both platforms:
+
+* accessibilityCompoundComponent: replace accessible-- indicates to assistive technologies that all the components inside this one should be treated as a single component.
+* accessibilityRelationships: describes how this component is related to others
```
necolas: I have a proposal internally that is based on Web's ARIA properties: `accessibilityLabelID` (equivalent to `aria-labelledby`), `accessibilityDescriptionID` (equivalent to `aria-describedby`), and `accessibilityFocusID` (equivalent to `aria-activedescendent`). All these props would take the value of the `nativeID` of the element they are associated with.

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+* accessibilityRange: expose information about range-based components such as sliders
```
necolas: I think we can probably learn from web here and expose the range props to the underlying accessibility layer of the platform. Or model this off ARIA props for progressbar etc

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+This property is already supported on Android, so no changes in Android support are required.
+
+#### iOS
+
+We believe this functionality can be implemented using accessibility notifications on iOS. For assertive live regions, the updated text of the component can be sent immediately to iOS as an announcement notification. For polite live regions, code can wait for the AnnouncementDidFinish notification before posting the updated text as an announcement notification.
+
+### accessibilityCompoundComponent
+
+We think the current implementation, which is called "accessible", should be renamed to accessibilityCompoundComponent, to more accurately represent what it does. The functionality should remain the same.
```
necolas: The behaviour of `accessible` has always felt a bit ambiguous to me, as on Android and Web it's equivalent to `focusable`. I don't think `accessibilityCompoundComponent` is any clearer a name than `accessible`

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+### accessibilityRelationships
+
+This property describes how this component is related to others. Relationship types include:
+
+* labelFor: This component labels the specified component
+* labeledBy: this component is labeled by the specified component
+* describedBy: this component is described by the specified components.
+
+When referring to other components, we may need to add an additional property to view-- namely accessibilityId. We need to find the best forward-looking approach for referencing other components.
```
necolas: We already have `nativeID` that we can use for this

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### iOS
+
+We believe this functionality can be implemented using accessibility notifications on iOS. For assertive live regions, the updated text of the component can be sent immediately to iOS as an announcement notification. For polite live regions, code can wait for the AnnouncementDidFinish notification before posting the updated text as an announcement notification.
+
+### accessibilityCompoundComponent
+
+We think the current implementation, which is called "accessible", should be renamed to accessibilityCompoundComponent, to more accurately represent what it does. The functionality should remain the same.
+
+### accessibilityRelationships
```
necolas: What does this API look like in practice? Is the value an object?

necolas:
For context, Twitter uses React Native to implement their PWA. The RN abstractions (which have web DNA) are considerably better for building modern web apps that straight DOM abstractions. However, Twitter relies on an undocumented feature of React Native for Web that passes `aria-*` props to the underlying DOM node to work around the missing accessibility features in React Native. Your proposals cover almost exactly the same problems that Twitter has from a web context and that I started bringing up at Facebook. So I'm really happy to see that Amazon is interested in improving the native accessibility story and would like to help make sure that the API maps cleanly to the web's comprehensive accessibility features (which influenced the introduction of `accessibilityRole` and `accessibilityStates`).

proposals/0000-a11y-cross-platform.md
marcmulcahy: Perhaps something like ARIA hiden would be adequate. Android's importantForAccessibility is a bad model to emulate-- it includes yes, no, and auto as values. Auto is the default I believe, and from the developer's point of view, is hard to use, since the rules for when something is automatically marked as unimportant for accessibility aren't clear. I think a better model is to assume everything is important for accessibility, unless explicitly marked otherwise. So it seems like some notion of hidden is adequate.

proposals/0000-a11y-cross-platform.md
marcmulcahy: We could standardize on ARIA's notion of a live region, and then dumb down the implementation to the various platforms. Android's notion of a live region is pretty basic (polite or assertive). iOS has no notion of live regions as far as I know, so we'd have to use iOS primitives to implement a subset of live region functionality.

proposals/0000-a11y-cross-platform.md
marcmulcahy: Makes sense. From naming point of view, are you in favor of using the ARIA namespace, or creating our own properties based on ARIA?

proposals/0000-a11y-cross-platform.md
marcmulcahy: We've found that in practice, exposing range information in a consumable way is fairly difficult. Exposing a minimum, current, and maximum value allow you to speak a percentage, which is fine for some types of progress bars. But for others, where the actual value is relevant (say a color contrast setting, stereo balance control, movie timeline slider, etc.) it would be nice to include some information that could cause the screen reader to speak either a string or the current value instead of a percentage. Slider's also imply programatic actions to adjust them.

proposals/0000-a11y-cross-platform.md
necolas: Do you think we should make `accessibilityElementsHidden` cross-platform? Mapping to `aria-hidden` on web and equivalent on Android?

proposals/0000-a11y-cross-platform.md
necolas: `accessibilityLiveRegion` is essentially already identical to ARIA live region, there's just a different in what the `none` state is called. So you could extend support for the existing API to iOS and then we'll have the 3 main platforms covered.

proposals/0000-a11y-cross-platform.md
necolas: What do you mean by "ARIA namespace"? Something like `accessibilityLabelledBy` vs `accessibilityLabelID`?

proposals/0000-a11y-cross-platform.md
necolas: > But for others, where the actual value is relevant (say a color contrast setting, stereo balance control, movie timeline slider, etc.) it would be nice to include some information that could cause the screen reader to speak either a string or the current value instead of a percentage.

ARIA has a property for that `aria-valuetext` - https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-2.html

I do like how understandable and extensible the `accessibilityRelationships` and `accessibilityRange` props names are vs having many more individual props ARIA.

Are you thinking it would look something like this (property names tbh for relationships)?

```
<View
  accessibilityRange={{ min, max, current, text }}
  accessibilityRelationships={{ labelID, descriptionID, focusID }}
/>
```

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+Several properties are missing from both platforms:
+
+* accessibilityCompoundComponent: replace accessible-- indicates to assistive technologies that all the components inside this one should be treated as a single component.
+* accessibilityRelationships: describes how this component is related to others
+* accessibilityRange: expose information about range-based components such as sliders
+
+## Motivation
+
+All accessibility properties should be available on all supported platforms. This reduces the burden on developers to understand the nuances of accessibility support on various platforms. It also allows JavaScript components to expose rich accessibility information, rather than relying on the accessibility support in native components.
```
necolas: 👍

proposals/0000-a11y-cross-platform.md
necolas: > Maybe I should drop this from this proposal, and make a separate proposal covering compound component handling?

I think that sounds good. The idea of an accessibility group is something that's a little complicated on web too, because you have both screen-reader accessibility and keyboard accessibility.

proposals/0000-a11y-cross-platform.md
necolas: Looks like it was introduced for both platforms here in June 2017 https://github.com/facebook/react-native/commit/70e04555226fe3a57eddfb7d601ad9ff600e2b40

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+### accessibilityCompoundComponent
+
+We think the current implementation, which is called "accessible", should be renamed to accessibilityCompoundComponent, to more accurately represent what it does. The functionality should remain the same.
+
+### accessibilityRelationships
+
+This property describes how this component is related to others. Relationship types include:
+
+* labelFor: This component labels the specified component
```
necolas: Is there a use-case you had in mind for having the relationship be defined in either direction like this?

proposals/0000-a11y-cross-platform.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Adoption strategy
+
+We don't believe this to be a breaking change. Developers who currently use these properties on Android or iOS can continue to do so, and remove conditional use to allow functionality on other platforms.
+
+## How we teach this
+
+The same documentation that is used to describe the properties on either iOS or Android can be marked as cross-platform for use on either. For the accessibilityRelationships property, labeledBy and labelFor are common paradigms across accessibility support on multiple platforms including web.
+
+## Unresolved questions
```
necolas: I have a related question (and [issue](https://github.com/necolas/react-native-web/issues/1181)) about content that is hidden visually but visible to screen readers. Typically it will be hidden headings so they show up in the screen reader headings list (e.g., mobile.twitter.com does this) and it is visually hidden with a particular CSS incantation. Would you / how would you do something like this in native apps?

proposals/0000-a11y-cross-platform.md
harinikmsft: +1 on the accessibilityLiveRegion property being more universal. We just implemented this for Windows (see PR [here](https://github.com/microsoft/react-native-windows/pull/2591)).

LisaPritchard: +1 for accessibilityLiveRegion on iOS

proposals/0000-a11y-cross-platform.md
jguddas: Maybe something like this?
```tsx
import React, { useEffect, useState } from 'react';
import { AccessibilityInfo, Text } from 'react-native';

const useIsScreenReaderEnabled = () => {
  const [isScreenReaderEnabled, setIsScreenReaderEnabled] = useState(false);

  useEffect(() => {
    AccessibilityInfo.addEventListener('screenReaderChanged', setIsScreenReaderEnabled);
    AccessibilityInfo.isScreenReaderEnabled().then(setIsScreenReaderEnabled);
    return () => AccessibilityInfo.removeEventListener('screenReaderChanged', setIsScreenReaderEnabled);
  }, []);

  return isScreenReaderEnabled;
};

const Example = () => {
  const isScreenReaderEnabled = useIsScreenReaderEnabled();
  return <Text>I {isScreenReaderEnabled ? 'love' : '❤️'} React Native for Web</Text>;
};
```

necolas:
I'm going to close this proposal as we've subsequently focused on cross-platform APIs based on the existing ARIA spec as per https://github.com/react-native-community/discussions-and-proposals/pull/496

PR #80, opened 2019-01-08
Proposal to access to Android's isInMultiWindowMode and isInPictureInPictureMode

sidferreira:
Since Nougat ( API Level 25 ) Android has both `Activity.isInMultiWindowMode` and `Activity.isInPictureInPictureMode`. Properties that so far I couldn't find ways to be aware if the app is in window mode or not. The idea is to offer such information to the developers in order to achieve better experiences with their apps.

### DIFF
--- a/proposals/0000-Access-to-Android-s-InMultiWindowMode.md
+++ b/proposals/0000-Access-to-Android-s-InMultiWindowMode.md
@@ -1,0 +1,53 @@
+---
+title: Access to Android's InMultiWindowMode
+author:
+- Sidney Ferreira
+date: 2019-01-09
+---
+
+# RFC0000: Access to Android's InMultiWindowMode
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


axemclion:
Does this have to be in core react native, or could this simply be a native module ? What support does it need from core react native ?

kelset:
Isn’t this more of a feature request than a proposal? 🤔

sidferreira:
@kelset yes, but I know RN team have a LOT of stuff to do. So, as it is not a BIG change, I wanted to both suggest the feature AND offer my self to do it(1). And this is the path I found to check if a PR has chances to be accepted.



1: I had a small repo that ended having a lot of people asking for maintenance and features. What did I learn? `Open Source Developers` are **NOT** my employees. So, if there's something I want I have two options: `a) suggest and wait` or `b) suggest and offer my self to do it`.

kelset:
Yeah absolutely, I mean - I share your experience 🤗

Simply, given that it is a “small” and precise feature request I think that the correct approach would be to just go for it and make the PR ;)

sidferreira:
@kelset Well if you are a collaborator and say so, well, I'll do it 👍 
Thanks for your time!

PR #86, opened 2019-01-21
Fix the numbers of RFCs

fson:
We had two of RFC0002 and two RFC0003's. I've changed the numbers to be unique.

screenshot 2019-01-21 at 13 04 59


cpojer:
Thanks. I think we should use pull request number instead of manual management of incrementing ids, tbh.

jamonholmgren:
I agree about PR number being the id of record.

kelset:
yeah I think we had that system because we copied some other RFC procedure but agree with this PR approach 👍

PR #103, opened 2019-02-19
Fix filename of November 2018 meeting

casperboone:
The filename currently says 2019, while the meeting took place in 2018. This PR changes the filename to include the correct year.


kelset:
Thanks for the PR @casperboone !

PR #110, opened 2019-03-15
Add possibility to trigger animation on touchable events bypassing js thread.

vaukalak:
Unfortunately js thread and the bridge are still bottleneck for some of animations, this proposal attempts to fix that for touchable-driven animations.

### DIFF
--- a/proposals/0000-press-animations-bypass-js.md
+++ b/proposals/0000-press-animations-bypass-js.md
@@ -1,0 +1,117 @@
+---
+title: Add possibility to onPressIn/onPressOut/onPress callbacks to trigger animation bypassing js thread.
+author:
+- Mikhail Boutylin
+date: 03-15-2019
+---
+
+# RFC0000: Add possibility to onPressIn/onPressOut/onPress callbacks to trigger animation bypassing js thread.
+
+## Summary
+
+Allow to pass `AnimationConfig` and it subinstances as `onPressIn`/`onPressOut`/`onPress` properties, So the animation can be triggered directly by the UI thread.
+
+## Basic example
+
+```
+/**
+ * @format
+ * @flow
+ */
+
+import React, {Component} from 'react';
+import {StyleSheet, Animated, Text, View, TouchableWithoutFeedback} from 'react-native';
+
+
+type Props = {};
+
+const animation = new Animated.Value(1);
+
+const animateIn = () => Animated.timing(
+  animation,
+  {
+    toValue: 2,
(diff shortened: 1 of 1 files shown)


patrickkempff:
I like this approach! 

I am wondering how the duration of the pressIn animation should be determined? Is it by using the `duration` param of the animation? or as long as the user holds the touchable? 

Let's say we use duration param of the animation itself, what happens to the animation when use user releases the touchable? does it cancel the  animation and starts the pressOut animation? How to make those seamless?

Also does passing animations to pressIn/pressOut only work with TouchableWithoutFeedback? Or also with the others like TouchableOpacity and TouchableNativeFeedback? What happens to the existing animations of those components?

vaukalak:
@patrickkempff so as we pass `AnimationConfig`, the duration is already there. And yes, animation will be stopped, when button has been released. I think `TouchableOpacity` and `TouchableNativeFeedback` will benefit of this api, cause their animations will be also more performant when using this approach. Anyway this brings me to think that maybe `pressIn` and `pressInAnimation` should be probably different properties, otherwise it will be impossibe to subscribe to touchable opacity `pressIn`.

brentvatne:
i don't believe that this is possible with the existing gesture system because almost all of the logic is performed in js. you can build this already today by using react-native-gesture-handler and react-native-reanimated though, and then build the same api you're proposing on top of that.

vaukalak:
@brentvatne so you mean, that, when I have code like
```
const scrollY = new Animated.Value(0);
...
<ScrollView
    style={styles.fill}
    scrollEventThrottle={16}
    onScroll={Animated.event(
      [{nativeEvent: {contentOffset: {y: scrollY}}}]
    )}
  >
```
All updates of scrollY internal value will still happen on js?

brentvatne:
well yes in this case it will because you didn't add `useNativeDriver` to the `Animated.event` - but also `ScrollView` is different from `Touchable*` because its gestures are managed by native, which makes this possible. `onPress` on `Touchable*` originates from JS code that processes touch events, whereas `onScroll` originates from native and is later dispatched as an event to JS.

brentvatne:
i'd direct efforts towards this issue instead: https://github.com/kmagiera/react-native-gesture-handler/issues/358

it's possible to implement this type of behavior right now using other gesture-handlers made available through this pkg, and will be quite easy to build something like the touchable api you mentioned above once the issue is resolved.

vaukalak:
@brentvatne thanks for the feedback.

PR #116, opened 2019-04-02
Add Peril to the Org RFC

orta:
> Use [Peril](https://github.com/danger/peril) to ensure consistency across all community React Native repos. This would allow the entire community to
operate under the same rules in PRs, and can provide a way to ensure future projects have the right infra baseline.

### DIFF
--- a/proposals/0006-add-peril-community.md
+++ b/proposals/0006-add-peril-community.md
@@ -1,0 +1,55 @@
+---
+title: Use Peril on React Native Community
+author:
+- Orta Therox
+date: 2019-1-4
+---
+
+# RFC0006: Use Peril on React Native Community
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


orta:
Ha, not really! It's kinda up to the community, Peril is a blank slate - I don't actively maintain any of these repos, so I'm not certain what people want. I threw in some examples into the RFC though. 

That said, here's what we do at Artsy: https://github.com/artsy/README/blob/master/culture/peril.md

Any of that is feasible for example.

kelset:
Uhm I feel that this conversation would belong better in the meta repo, it's not really about React Native per se -> https://github.com/react-native-community/meta

I've already linked it to a discussion which may be [related](https://github.com/react-native-community/meta/issues/8), but I feel that maybe it can fit in more than one?

orta:
Yeah, that makes sense 👍

Issue #126, opened 2019-05-13
Platform specific colors via PlatformColor


turnrye:
**Edit: A more general proposal has been posted by @tom-un to support generic platform defined colors. [Jump to that proposal](https://github.com/react-native-community/discussions-and-proposals/issues/126#issuecomment-565913027). That proposal now supersedes the one in this top post**.

<details><summary>Original Proposal</summary>
<p>
### Introduction

Android Q introduced a [Dark theme](https://developer.android.com/preview/features/darktheme) across the system and applications. From their presentation, ["Users will expect apps to have a dark theme"](https://youtu.be/OCHEjeLC_UY?t=1327). The way we do styling on RN today isn't directly compatible with their solution, but it's close.

### The Core of It

I opened https://github.com/facebook/react-native/pull/24790 with a proposed solution, but it admittedly had some issues:

- It was on `Platform`, but it's more prone to change than the rest of those values
- It closely followed Android's implementation, but doesn't anticipate what iOS or out-of-tree platforms might use
- Android normally recreates the activities when this value changes, but that isn't a safe way to lead to re-rendering (which would be common if the user has enabled it by time-of-day or as part of battery saver)

### Discussion Points

- What's a better home for this than `Platform`? While it is a device setting, it's not quite in the same class as things like `Version`.
- Should we stick with the interface that we know (Android's) or anticipate a more universal solution using perhaps https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
</p>
</details>

acoates-ms:
@tom-un, might be able to fill in some more details.  But for our macOS RN platform we extended the valid values for color.  So you can specify either system colors, or be able to specify colors that change based on the native theme Something like:

<View background={{dynamic:{light: 'red', dark:'pink'}}} /> 
<View background={semantic: "windowBackgroundColor"} /> 

I believe there was some plan to look at contributing that back for iOS too.

mmmulani:
thanks for starting the proposal!

> Android normally recreates the activities when this value changes, but that isn't a safe way to lead to re-rendering (which would be common if the user has enabled it by time-of-day or as part of battery saver)

from my reading of the docs, it seem like this changed and depending on how it's implemented, activities might not be recreated when the system theme changed. Because of that, I think a pattern similar to how we treat orientation makes more sense

turnrye:
[...]

PR #162, opened 2019-10-17
Create 0000-apptheme-apis.md

kikisaints:
This proposal is a request to add theme changed event that detects when the system's theme goes from dark to light or light to dark.

This api feature addresses the [2206](https://github.com/microsoft/react-native-windows/issues/2206) issue in the react-native-windows repo and has already been implemented in that separate plugin.

Incorporating this change into the React Native repo would actually be a PR from the react-native-windows repo and minor API changes where necessary. This proposal details out potential final React Native api names.

### DIFF
--- a/proposals/0000-apptheme-apis.md
+++ b/proposals/0000-apptheme-apis.md
@@ -1,0 +1,95 @@
+# Responding to Light/Dark Themes in React Native
+
+## Summary
+With the introduction of React Native for Windows, there needs to be a way to utilize/work with the Light and Dark themes at the JavaScript layer.
+
+>This feature was pitched and integrated into the [React Native for Windows plugin in the PR here.](https://github.com/microsoft/react-native-windows/pull/2684)
+
+## Motivation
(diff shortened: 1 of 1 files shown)


dimitarnestorov:
So to get the string `dark` or `light` you're proposing that people write `AppThemeState.currentTheme.currentTheme`?

kikisaints:
> So to get the string `dark` or `light` you're proposing that people write `AppThemeState.currentTheme.currentTheme`?

That was not my intention. I was thinking the `AppThemeState.currentTheme` would hold the string values either "light" or "dark". I can provide an extra example in the proposal if that's not clear?

kikisaints:
@kelset  Looks like you already have an appearance module coming in for iOS/Android that may render this PR obsolete. Can you clarify if it is indeed solving the intention of this proposal?

kelset:
Hey @kikisaints thanks for the ping - sadly I have no idea as I wasn't directly involved in this. You've already commented in https://github.com/react-native-community/discussions-and-proposals/issues/126 so I know you are aware of that issue, sorry but I don't know any more than what has been written there 😓

PR #165, opened 2019-10-21
Configure nsurl session

hakonk:
This proposal addresses how the HTTP session configuration of `RCTHTTPRequestHandler` can be handled.

### DIFF
--- a/proposals/0000-Configure-NSURLSession.md
+++ b/proposals/0000-Configure-NSURLSession.md
@@ -142,7 +142,7 @@ static NSURLSessionConfigurationProvider theProvider;
 @end
 ```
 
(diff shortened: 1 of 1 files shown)


hakonk:
As pointed out by @rjerue in #166, [the Android implementation](https://github.com/facebook/react-native/commit/22efd95be1f0b236eeaaa8a8e6d01e89771c9543) allows for setting the HTTP client factory with a static method. The way I see it, that is similar to the alternative approach listed in this RFC. I believe it's beneficial that the customization mechanism is similar across platforms. Thus, perhaps that is the best approach on iOS as well. Curious to hear what your thoughts are!

TheSavior:
I think the discussion of this would be best on a PR to the main repo with the actual code. I think this since this is focused on a specific change it will be easier for you to receive feedback that way than here on this repo.

Issue #96, opened 2019-01-28
Native Modules


grabbou:
I've been thinking about the state of native modules and what we could do better to improve them. After working with few other React Native team members on the details, I think I am ready to share what I've came up with so far. 

This issue is based on some feedback I gathered from https://github.com/react-native-community/discussions-and-proposals/issues/60. Since the scope is a bit broader than just linking libraries, I decided to open a new issue.

This document describes the current developer experience working with native modules, most important problems and areas where we could improve. The intent is for this issue to become umbrella for various RFCs that will be submitted as a part of it.

## Glossary of terms

I want this issue to be super easy to understand, hence I decided to break down the terminology that I use throughout this issue. If you find anything unclear, please comment and I'll add it here.

- **React Native library** - a npm package that can contain one or many native modules
- **Native Module** - a module that extends capabilities of React Native. It usually ships with Android  and iOS source code. 
- **Native Dependency** - native module that depends on other Java and/or ObjC libraries, that may be distributed as source, or as compiled binaries.
- **Linking a dependency** - the process of adding files present in a native module to your iOS and Android projects
- **`react-native`**- React Native CLI
- **`react-native link`** - command to link a dependency to your project. It is a part of React Native CLI

## Problem

Working with native modules in React Native can be overwhelming. From the library author perspective, it is not clear how native dependencies could be handled and what is the recommended structure for a module. 

From the app developer perspective, it is not easy to consume React Native dependencies from the npm registry. Some of them are pure Javascript libraries that can be installed just like any other Javascript dependency. Others require additional steps to be performed (namely Android and iOS files to be added to respective iOS and Android projects). This can be confusing and increases the entry-barrier for the developers that want to extend basic capabilities of React Native with 3rd party modules.     

See the reasons highlighted below for details.

### Lack of standard

There is no standard or recommended way to create a React Native library. This problem is especially visible on iOS, where no built-in package manager for native dependencies is available (see "No support for native dependencies on iOS" section). 

As a result, each React Native library tends to have a slightly different structure in order to satisfy their use-case. This makes it hard to provide a decent tooling support that can automate most of the repeated steps (see "Poor tooling support" section).

It is also not clear at a glance what steps need to be performed. This degrades the developer experience from a trivial “yarn add” to many complex instructions.
 
### No support for native dependencies on iOS

Some React Native dependencies rely on native libraries in order to provide their functionality. It is a common practice to wrap a native SDK and expose it to React Native developers. 

[...]

PR #204, opened 2020-02-11
React Native Test App as a Package

tido64:
We're proposing to provide React Native test apps for all platforms (currently Android, iOS, Windows and macOS) as a package. Our aim is to make it easier for package maintainers to add support for a platform, to be able to switch between React Native versions, run on CI, etc. The package should take care of the native integration so that developers can spend more time on their component/feature.

You can find a rough (but working) prototype of this here: https://github.com/microsoft/react-native-test-app. We are consuming this internally and are actively iterating towards the goals detailed in the proposal.

- Android: @ArazAbishov, @sweggersen
- Windows: @acoates-ms, @afoxman
- macOS: @alloy, @tom-un

cc @alexpomsft, @danrydholm, @khovik

### DIFF
--- a/proposals/0000-React-Native-Test-App-as-a-Package.md
+++ b/proposals/0000-React-Native-Test-App-as-a-Package.md
@@ -1,0 +1,285 @@
+---
+title: React Native Test App as a Package
+author:
+  - Tommy Nguyen
+date: February 2020
+---
+
+# RFC0000: React Native Test App as a Package
+
+## Summary
+
+Provide a React Native test app for all platforms as a package.
+
+A working prototype of this can be found here:
+https://github.com/microsoft/react-native-test-app
+
+## Motivation
+
+Maintaining and keeping React Native up to date with the latest version can be
+time consuming. Things can break if you're not careful, even with
+[React Native upgrade helper](https://react-native-community.github.io/upgrade-helper/).
+Not all developers have much experience in native development on a single
+platform, let alone _all_ platforms they want to support. Adding support for
+another platform, such as Windows or macOS, can be overwhelming.
+
+Consider the following scenarios:
+
+- As maintainers of a community package in our spare time, we don't have the
+  capacity to create and maintain test apps.
+
+  - Our inner development loop consists of using `npx react-native init`, and
+    testing things manually.
+  - Alternatively, we have set up CI for iOS and Android, and would love to add
+    support for Windows and macOS but lack the resources.
+
+- We have a monorepo with multiple separate experiences that all need test apps.
+  It is a pain having to upgrade all of the test apps individually with every
+  release. Our upgrade process can sometimes take weeks to perform.
+
+- As a consumer of React Native, we should try out release candidates to find
+  issues early and contribute back, but due to one of the scenarios above, there
+  is simply too much friction.
+
+The proposed package will, at minimum, provide a customizable test app for all
+the major platforms that React Native currently runs on, i.e. iOS, Android,
+Windows, and macOS. The ultimate goal of this package is to make adding support
+for a platform as simple as `npm add --save-dev`, and upgrading to latest React
+Native is just a version bump.
+
+We expect the package will:
+
+- Help developers set up a consistent native environment for both local
+  development and CI agents
+- Reduce or eliminate the friction to upgrade React Native, and try out release
+  candidates
+- Reduce or eliminate the overhead of adding support for new platforms
+
+With common test apps, we will also be able to provide additional value for
+everyone simultanously. For instance:
+
+- We will handle upgrading of the test apps to support the latest mobile SDK
+- We can provide a common infrastructure for native testing
+- We can provide optional modules for commonly used services, such as auth or
+  App Center integration
+
+## Detailed design
+
+Given an existing React Native app/library, we imagine the ideal flow for
+**initial integration** with the native test apps should be:
+
+1. Create folder/package for the test app to live in.
+   - This package should have a dependency on the app/library to be tested, and
+     on the test app package.
+2. Create a manifest to declare entry points. Optionally, it should also be
+   possible to declare initial props, native source files and resources.
+   - This is a **one-time cost**, and should be stable across versions.
+   - If there is a single entry point, the app may boot directly into this app.
+   - Otherwise, if there are multiple entry points, we will show them in a list.
+     - It should still be possible to have the app boot into one of these via
+       deep linking.
+   - Native dependencies should be taken care of by auto-linking behind the
+     scenes.
+3. Run `npm install` inside this folder as one would usually do.
+4. Depending on the desired target platforms, the next steps are:
+   - Android:
+     - Make minimal changes to `build.gradle` (one-time cost)
+     - Open the project in Android Studio
+   - iOS:
+     - Make minimal changes to `Podfile` (one-time cost)
+     - Run `pod install`
+     - Open the newly generated Xcode workspace
+   - Windows:
+     - <@afoxman or @acoates-ms to fill in>
+   - macOS:
+     - See steps for iOS
+
+A hypothetical package should end up looking something like below:
+
+```
+my-awesome-component
+├── example
+│   ├── App.js
+│   ├── build.gradle (new/modified)
+│   ├── package.json
+│   ├── Podfile (new/modified)
+│   └── yarn.lock
+├── my-awesome-component.podspec
+├── my-awesome-component-dev.podspec (new, optional)
+├── package.json
+├── src
+└── yarn.lock
+```
+
+For a hypothetical monorepo:
+
+```
+my-monorepo
+├── package.json
+├── packages
+│   ├── my-awesome-component-android
+│   │   ├── build.gradle (new/modified)
+│   │   ├── package.json
+│   │   └── src
+│   ├── my-awesome-component-core
+│   │   ├── package.json
+│   │   └── ...
+│   └── my-awesome-component-ios
+│       ├── my-awesome-component.podspec
+│       ├── my-awesome-component-dev.podspec (new, optional)
+│       ├── package.json
+│       ├── Podfile (new/modified)
+│       └── src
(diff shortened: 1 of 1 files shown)


kelset:
cc @bartolkaruza

TheSavior:
I love this. I think making it easier for module owners to verify their modules works on different versions will be great.

I'm a little nervous about this part
> Aim to become part of the default create-react-native-app template

The scenario for people who maintain apps to test against other versions seems like a bit of a different problem to me. For example, they probably don't need to make sure their app works across multiple versions of React Native. 

Do you see it as important to support app developers with this tool? Or does it make sense to have it be focused on library authors / platform authors?

proposals/0000-React-Native-Test-App-as-a-Package.md
```diff
@@ -0,0 +1,285 @@
...
+solution to get started, and you don't mind that it's including a lot of things
+that your app won't use. Once your app matures to a point where it needs better
+control over what's included, or it needs to consume native modules that aren't
+bundled in the SDK. Ejecting brings you back to the world of having to maintain
+React Native yourself.
+
+Expo also isn't aimed at core contributors/maintainers. Its main goal is to
+simplify getting started with your React Native app. If you're looking to
+implement a native module of your own, you'll need to eject and stay ejected.
+This defeats the purpose of using Expo in the first place.
```
brentvatne: I'd replace this entire section on Expo with the following:

> [Expo](https://expo.io) provides an app called Expo client that includes multiple versions of the runtime in a single binary and makes it easy to switch between those versions. Each version of the runtime uses a single version React Native. [Snack](https://snack.expo.io) is an in-browser editor like [CodeSandbox](https://codesandbox.io) for React Native apps, and it allows you switch between the React Native version that an app runs against by selecting a different SDK versions in a dropdown menu. SDK versions usually map to distinct React Native versions, eg: SDK 36 uses React Native 0.61 and SDK 35 uses React Native 0.59.
>
> This is the quickest and easiest way to test React Native libraries for iOS, Android, and web, provided that they depend on primitives in React Native core or the most common dependencies in the ecosystem like react-native-gesture-handler, react-native-reanimated, react-native-maps, react-native-svg, and so on.
>
> There are two reasons why this isn't feasible for the particular use case described in this proposal. First, as mentioned already, a library includes its own native dependencies it cannot be run inside of Expo client unless the appropriate native code is included in the client at build-time. Second, Expo client apps are not yet built for Windows and macOS, and so you would not be able to test on those platforms using it.

brentvatne:
One thing that is notably missing from the list of platforms that this should support is web. Do you have any interest in building this into the package as well or do you think this is out of scope @tido64?

tido64:
> @TheSavior: Do you see it as important to support app developers with this tool? Or does it make sense to have it be focused on library authors / platform authors?

I think it completely makes sense to have it be focused on library/platform authors. I guess what I was trying to say is that we want a simple way to get started with a completely new component/library with support for all platforms right off the bat. Maybe a library template for `react-native-cli` is what I should have put here. E.g. a developer could run `react-native init ProjectName --template "react-native-library-template"`, and you'd get a project with test apps for all platforms ready to go. Does that make more sense?

> @brentvatne: One thing that is notably missing from the list of platforms that this should support is web. Do you have any interest in building this into the package as well or do you think this is out of scope @tido64?

Our primary focus will be to support mobile and desktop platforms. That is what we currently have capacity for. @acoates-ms correct me if I'm wrong. That said, I don't think anyone would be opposed to getting help with additional platforms. In fact, we'd welcome any contributions.

proposals/0000-React-Native-Test-App-as-a-Package.md
```diff
@@ -0,0 +1,285 @@
...
+uses today, and we can extend it to support our scenarios.
+
+For iOS (and macOS), we will need to extend the format to include additional
+native source files and assets. An Xcode project generated today, will include a
+main app target, and two test targets: One for "normal" tests, and one for UI
+tests. We will need to be able to generate `.podspec` files that can support
+all.
+
+Entry points will also be declared in the manifest. Entry points tell the test
+app how to launch your app. For instance, given the declaration below:
```
grabbou: I am wondering if the concept of `entry points` is necessary here. Wouldn't the end result be the same if `AppRegistry.registerComponent` was used directly in the source code and then, the loader just [accessed the registry](https://github.com/facebook/react-native/blob/master/Libraries/ReactNative/AppRegistry.js#L165) and read all the runnables?

I would stick to a single JavaScript file as an entry point where all the components are registered.

proposals/0000-React-Native-Test-App-as-a-Package.md
```diff
@@ -0,0 +1,285 @@
...
+│   └── my-awesome-component-ios
+│       ├── my-awesome-component.podspec
+│       ├── my-awesome-component-dev.podspec (new, optional)
+│       ├── package.json
+│       ├── Podfile (new/modified)
+│       └── src
+└── yarn.lock
+```
+
+Platform specific project files should no longer be checked in. They are either
```
grabbou: There were a lot of discussions about "generating native files at a build time" in the past, unfortunately, all of them ended up being stuck after a while.

You may find this issue https://github.com/facebook/react-native/issues/6292 useful for some additional context.

grabbou:
On a high level, this proposal seems like an ability to create apps (production or test apps) by writing JavaScript-only files and abstracting away the native code. Is there anything that makes this proposal limited to creating test apps only? If not, then, I think there's an interesting opportunity here.

Either way, I have left some comments here and there to accelerate further discussions. Really looking forward to the next steps as this proposal seems like something that we have been trying to implement for a while already.

@tido64 

> E.g. a developer could run react-native init ProjectName --template "react-native-library-template", and you'd get a project with test apps for all platforms ready to go. Does that make more sense?

In fact, this idea was floating around for a while already. For example, in this issue, https://github.com/react-native-community/discussions-and-proposals/issues/96 that was created as a motivation for `autolinking`, one of the next steps was to create a good boilerplate for libraries that use Yarn workspaces and provides ready-to-use test application.

I think the outcome of this proposal could be a creation of such a template. I am more than happy to help out with this part.

TheSavior:
@tido64, ah. I understand better now. If you want to make this a default part of creating new modules, I don't think react-native init is the right place as that is mostly known for creating end user apps. After a quick google search, I found these two projects that seem to be for helping bootstrap a new native module / component: [bob](https://github.com/react-native-community/bob), and [creat-react-native-module](https://github.com/brodybits/create-react-native-module).

proposals/0000-React-Native-Test-App-as-a-Package.md
tido64: `entryPoint` is more for packages that provide "screens" that are directly accessible from native (e.g. in a brownfield/hybrid app).

What you're suggesting makes a lot of sense for packages that provide components. Because they aren't apps, it's likely that the JS test app is the only registered component. There are no reasons we couldn't have both. We can read from the registry if no entry points were declared in the manifest. I'll make sure to add this in the next iteration.

tido64:
> @grabbou: On a high level, this proposal seems like an ability to create apps (production or test apps) by writing JavaScript-only files and abstracting away the native code. Is there anything that makes this proposal limited to creating test apps only? If not, then, I think there's an interesting opportunity here.

There shouldn't be? As far as I can see, you will have to add support for declaring app name and bundle id etc. in the manifest. If that's something you want to work on, you should go for it. That said, I think our top priority should be an excellent dev experience.

> @grabbou: I think the outcome of this proposal could be a creation of such a template. I am more than happy to help out with this part.

I'll hold you to your word 👍

tido64:
Hi folks,

Ever since we've published this PR, we've been improving the test apps and ironing out the wrinkles. We've added support for macOS, and Windows support is in progress. We've been dogfooding it for quite some time now, and would really appreciate it if you could try it out and provide feedback. The test app package is now published as [`react-native-test-app`](https://www.npmjs.com/package/react-native-test-app). All our documentation can be found in [the wiki](https://github.com/microsoft/react-native-test-app/wiki).

As always, feel free to file issue in [react-native-test-app/issues](https://github.com/microsoft/react-native-test-app/issues) or reach out directly with questions in the usual channels.

asklar:
> Hi folks,
> 
> Ever since we've published this PR, we've been improving the test apps and ironing out the wrinkles. We've added support for macOS, and Windows support is in progress. We've been dogfooding it for quite some time now, and would really appreciate it if you could try it out and provide feedback. The test app package is now published as [`react-native-test-app`](https://www.npmjs.com/package/react-native-test-app). All our documentation can be found in [the wiki](https://github.com/microsoft/react-native-test-app/wiki).
> 
> As always, feel free to file issues in [react-native-test-app/issues](https://github.com/microsoft/react-native-test-app/issues) or reach out directly with questions in the usual channels.

<RNWindows hat on>
My main concern with this is it will make it pretty difficult to make changes to the rnw CLI without breaking apps/modules built with this tool. For example @jonthysell found that this tool places assets in locations that break certain assumptions that the RNW CLI makes which break autolinking.

tido64:
> My main concern with this is it will make it pretty difficult to make changes to the rnw CLI without breaking apps/modules built with this tool. For example @jonthysell found that this tool places assets in locations that break certain assumptions that the RNW CLI makes which break autolinking.

@asklar We'd be happy to make any changes that will be more future-proof. Please file an issue on the repo or reach out directly to us.

tido64:
Hi folks,

As of [v0.2.5](https://github.com/microsoft/react-native-test-app/releases/tag/0.2.5), we now support `react-native-windows` 0.62 and 0.63. This also means that we've caught up with the latest releases across all supported platforms, while also being backwards compatible to 0.60 on certain platforms. You can refer to the [React Native versions](https://github.com/microsoft/react-native-test-app/wiki#react-native-versions) table in our wiki for more details.

Please give it a go and let us know how you go. If you hit any issues, you can file it here: https://github.com/microsoft/react-native-test-app/issues

PR #219, opened 2020-04-21
Add DataKey Prop in SectionList

Rishabh-Rathod:
Adding a Prop to SectionList (i.e. dataKeyExtractor) can avoid data processing to modify data required as per it.
For now it only allows sections as  [{ data: [1,2,3] }, { data: [4,5,6]}
if data is like this  [{ items: [1,2,3] }, { items: [4,5,6]}
then it needs to be converted to  [{ data: [1,2,3] }, { data: [4,5,6]} it will be better to provide dataKeyExtractor where user can provide key of the array keeping by default key as "data"

### DIFF
--- a/proposals/0000-SectionList-dataKeyExtractor.md
+++ b/proposals/0000-SectionList-dataKeyExtractor.md
@@ -1,0 +1,80 @@
+---
+title: Adding a SectionList dataKeyExtractor prop
+author:
+- Rishabh
+date: 22/04/2020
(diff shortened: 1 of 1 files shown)


kelset:
👋 there - as stated in the README and the issue template, this repo is for long-form discussions. Yours seems more a precise feature request, so I suggest you open a PR against the main repo so that it can be discussed properly.

PR #329, opened 2021-02-01
Create 0000-change-official-slider-implementation.md

Sharcoux:

### DIFF
--- a/proposals/0000-change-official-slider-implementation.md
+++ b/proposals/0000-change-official-slider-implementation.md
@@ -14,7 +14,7 @@ In one year, the only point that seems to have moved on on this slider is the su
 
 ## Possible solution
 
-I Propose that we make @sharcoux/slider the new official implementation of this slider. Here is a list of why:
+I Propose that we make [@sharcoux/slider](https://www.npmjs.com/package/@sharcoux/slider) the new official implementation of this slider. Here is a list of why:
 
 - It is easy to maintain as the implementation is common for all devices
 - It depends only on react-native
--- a/proposals/0000-change-official-slider-implementation.md
+++ b/proposals/0000-change-official-slider-implementation.md
@@ -1,0 +1,25 @@
+# Improve the implementation of Slider
+
+## Problem
+
+The current 'official' implementation of slider in React Native, provided by [react-native-slider](https://github.com/callstack/react-native-slider) is full of very annoying bugs, making it unsatisfying.
+
+Here is a list of the more dramatic ones:
+
+- [Impossible to style the slider](https://github.com/callstack/react-native-slider/issues/254) or the [track](https://github.com/callstack/react-native-slider/issues/197), or the [thumb](https://github.com/callstack/react-native-slider/issues/169)
+- Range slider is not supported (see requests [here](https://github.com/callstack/react-native-slider/issues/228) and [here](https://github.com/callstack/react-native-slider/issues?q=is%3Aissue+is%3Aopen+range+slider)
+- There are plenty of bugs making it unreliable. Just take a look [at the list](https://github.com/callstack/react-native-slider/issues) and see that the issues are not just a bit annoying, but really critical
+
+In one year, the only point that seems to have moved on on this slider is the support for Windows. The support and the follow up on all the bugs is really poor.
+
+## Possible solution
+
+I Propose that we make @sharcoux/slider the new official implementation of this slider. Here is a list of why:
+
+- It is easy to maintain as the implementation is common for all devices
+- It depends only on react-native
+- It also brings support to range slider and many of the propositions made in the issues of react-native-slider (like stylings).
+- It is way lighter
+- It is easier to understand and fix if needed as there is no native code that require people knowing those languages.
+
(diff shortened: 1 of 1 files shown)


wojteg1337:
Well, the core idea of react-native-slider is to have native implementations (UISlider / AppCompatSeekBar) exposed to JS.

I'm pretty sure that any JS implementation won't replace current community implementation.

michalchudziak:
Hey @Sharcoux,

as one of the maintainers of react-native-slider I'd like to follow up on few points that you've outlined in your pull request.

First of all, I feel like your description of the issue uses a bit offensive language, which in my opinion does not facilitate constructive discussion. The current implementation of `react-native-slider` has been with us for years, and is a result of collaborative work of many developers. Yes, it does contain a few bugs or missing features (like most open source libraries) and we are more than happy to accept pull requests fixing them. 

The library is widely used (> 60k downloads weekly), and while the number of downloads doesn't mean that much per se, it implies the number of people that are continuously testing the solution and reporting bugs. Do you honestly think, that after reaching such number of users (which is 200x bigger), your solution won't need more customisation and won't have more edge-case bugs reported?

You mentioned us not actively developing library. I would argue with that, because we have added windows and web support, as well as accepted several pull requests for new features and bug fixes. Of course, we could commit more time to push forward some bugs, but it is not what Open Source is about - we are open to collaborate with others, and never turned down anyone, including you and your PRs.

Speaking of solutions, I do not think replacing current official implementation with the one you proposed is a way to go. In fact there are other mature JS-based solutions that could be considered for that case as well. We put a lot of effort to community work, and I agree, there's a lot of place for more work and continuous improvements. That's why I'd like to propose you joining our maintenance team for `react-native-slider`. I believe that with your motivation we could make this project even better and accessible for the larger audience!

Peace :)

Sharcoux:
@wojteg1337 I might be wrong, but I think that the core idea of react-native-slider should be to provide the best and most reliable and customizable slider feature possible to the users, independently of if how it is working under the hood, especially when the final code will, anyway, be transformed into a native code thanks to react-native.

@michalchudziak I'm sorry if you found my tone too aggressive. I'd gladly edit any sentence that you thought harmful.

The core issue here is that the current project has an enormous inertia, mainly because of the triple parts: web + android + iOS, and @michalchudziak I think that you can agree that you don't have time to maintain and improve those parts anymore. Relying on PR seems to not have moved us so far until now : [Adding thumb style](https://github.com/callstack/react-native-slider/issues/40) is asked since april 2019 and should be considered a basic feature when talking about a slider, and it's still not solved. And I don't say it to be mean. I totally understand that you have thousands of other things to take care of.

For me, passing to a full RN version and getting rid of the native part would make the community able to participate more and help the tool evolve faster.

Sharcoux:
I'm not trying to blame anyone or point finger or complain. I just think that there is an issue. A lot of people are asking for features that are nothing complicated and it's not being done, and I think that we should try to find causes and solutions.

For me, I think that the issue is that maintainers of this project fall in at least one of those 2 categories :
- Those who just don't have the time anymore to dedicate to this project and solve the issues
- Those who don't have the knowledge to mess in the native part 

This results in the project being very slow to improve even for simple feature requests or bug corrections.

@michalchudziak Do you think that am wrong about this? And if so, can you explain your analysis about the inertia of react-native-slider?

kelset:
Sounds like this conversation is about "official-ity" of a given repo vs the other, and I want to jump in to clarify something. Components that used to be in core, then got "lean-core'd" out have been migrated out of the React Native Community github org precisely because this aura of official-ity was a misunderstanding.

You can read more about that [here](https://github.com/react-native-community/discussions-and-proposals/blob/master/partners/0001-organization-repository-policy.md#react-native-community-will-transition-towards-hosting-only-the-repositories-that-fit-into-category-3).

So this whole discussion has no place to be; neither solution is nor will be the official one. They will be just two alternatives in the same field.

Now, I can see two reasons why you might perceive the slider implementation maintained by Callstack as "official":

1. The documentation redirects to it: https://reactnative.dev/docs/slider. ---> I would propose you do a PR to that page to rephrase that to something like "Deprecated. Consider using alternatives such as callstack/slider or <your_library> instead".
2. the Callstack npm package is under "@react-native-community" npm registry. This has been kept this way for the libraries migrated out of the GitHub org to avoid a pain for all developers to have to change their dependencies. If you feel strongly about this being an issue (and I can see why) I would propose you open an issue on this repo along the lines of "Stop using the @react-native-community npm org for packages not in the org" or something like that.

I hope this clarifies the situation.

PS: I agree with @michalchudziak that the tone used in this conversation is midly aggressive and makes harder to want to have conversations.

Sharcoux:
@kelset You totally understood my issue and I'll follow your advice. And to show you until where it extends, I have been told by some people that they prefer using the @react-native-community/react-native-slider because it is the official repository.

About the tone, I'm sorry but I don't see where I'm being aggressive. Could you highlight a sentence that tilted you? As I said, it is not my intention.

Maybe is it because I say that the project is having poor follow up on the bugs? But I really don't say this to blame anyone. I don't think that @michalchudziak would consider that he is still actively developing the project. If he is, indeed it can be considered aggressive, but it's easy to see that it is not the case. The maintainers are probably overwhelmed by other things to do and they don't have to feel ashamed about it.

Sharcoux:
@kelset Just to illustrate better the issue: (This is what happen)[https://github.com/facebook/react-native-website/pull/2469#issuecomment-771577785] when following your advice. People consider that @react-native-community packages are maintained by the React Native Community.

Sharcoux:
The discussion will continue [here](https://github.com/react-native-community/discussions-and-proposals/issues/330#issuecomment-772666380)

PR #335, opened 2021-02-27
Keyboard-friendly List Virtualization Behavior

NickGerleman:
VirtualizedList virtualizes away items based on proximity to the lists viewport,
without consideration for other features such as focus or selection. This
strategy limits the ability to correctly implement some common keyboard
interactions. This especially impacts desktop platforms (react-native-web,
react-native-windows, react-native-macos), but also impacts the tablet with
keyboard form-factor on iPadOS/Android.

This proposal outlines these issues, proposing a set of behavior changes and new
APIs to address them. It does not address concrete implementation.

### DIFF
--- a/proposals/0006-Keyboard-Friendly-Virtualization.md
+++ b/proposals/0006-Keyboard-Friendly-Virtualization.md
@@ -1,0 +1,186 @@
+---
+title: "RFC0006: Keyboard-friendly List Virtualization Behavior"
+author:
+  - Nick Gerleman
+date: February 2021
+---
+
+# RFC0006: Keyboard-friendly List Virtualization Behavior
+
+## Summary
+
+VirtualizedList virtualizes away items based on proximity to the lists viewport,
+without consideration for other features such as focus or selection. This
+strategy limits the ability to correctly implement some common keyboard
+interactions. This especially impacts desktop platforms (react-native-web,
+react-native-windows, react-native-macos), but also impacts the tablet with
+keyboard form-factor on iPadOS/Android.
+
+This proposal outlines these issues, proposing a set of behavior changes and new
+APIs to address them. It does not address concrete implementation.
+
+## Basic example
+
+```jsx
+<FlatList
+  ...otherProps,
+  keepAreasRealized={{
+    focused: false,
+    home: [true, {windowSize: 1}],
+  }}
+/>
+```
+
+## Motivation
+
+### Keyboarding issues with VirtualizedList (and FlatList/SectionList)
+
+Keyboard users expect to be able to navigate between between views inside of a
+list via tab-loop or arrow keys. A component with focus may move out of
+visibility by panning or scrolling a parent surface. The focused component may
+be virtualized away after leaving visibility.
+
+Removing components in the focus chain breaks some common keyboard scenarios:
+1. Moving to a previous/next selected item via arrow-key
+1. Tabbing focus to a user-definable next component
+1. Imperatively moving focus to a user-defined component
+
+## Detailed design
+
+### Realization windows
+
+Focus may be synchronously moved by the native UI framework. This forces us to
+prevent virtualization (keep realized) any components which focus may be moved
+to. We may see additional navigation after the initial focus change, unblocked
+on the rendering of new components. This necessitates a chain of additional
+realized components.
+
+Common keyboarding scenarios should "just work" without having to change
+settings. Three of these scenarios include:
+
+1. Navigating from one child component of a rendered item to another
+1. Arrowing between focused items
+1. Shifting to first/last focused item via home/end keys
+
+`VirtualizedList` has enough internal knowledge to transparently support these
+scenarios. This can be accomplished by adding three **"realization windows"**:
+1. **"focused"** Realize cells around a cell with active focus. Can be
+tracked by listening to bubbling onFocus in the default `CellRendererComponent`.
+1. **"home"** The area at the beginning of the list (for home key).
+1. **"end"** The area at the end of the list (for end key).
+
+The above scenarios are not exhaustive. E.g. we may want to keep a *selected*
+item realized that is not *focused*. These scenarios may require additional APIs
+allowing external users greater control over realization behavior. These could
+potentially be exposed in the future as dynamic/user-controllable realization
+windows, or separate high-level APIs.
+
+### Endpoint-specific default behaviors
+
+VirtualizedList should be keyboard-friendly by default on desktop platforms as
+they are oriented towards usage of a physical keyboard. This additional
+realization is not free however, increasing memory usage to retain additional
+cells. This tradeoff may not make sense for mobile.
+
+This leads to desired behavior of:
+- **Keep FlatList focus-safe on desktop:** Common realization windows for
+needed for keyboard navigation are enabled by default.
+- **Keep FlatList memory usage steady on mobile:** New realization windows
+are opt-in where not pay to play.
+
+The full matrix of behaviors is described in the *API Reference* below. In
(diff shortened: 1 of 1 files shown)


proposals/0006-Keyboard-Friendly-Virtualization.md
```diff
@@ -0,0 +1,185 @@
...
+   * https://reactnative.dev/docs/virtualizedlist#windowsize]
+   */
+  home?: RealizationWindow;
+
+  /**
+   * Keeps the end area of the list realized.
+   * - Enabled by default on desktop (for end-key navigation)
+   * - Disabled by default on mobile/TV
+   *
+   * Defaults to **half** the (VirtualizedList windowSize)[
```
NickGerleman, line 153: Half of windowsize means we have a full buffer at both ends, but also means we effectively double the number of realized cells on desktop. Think this should maybe be lower, but there are also some nice properties with current value.

proposals/0006-Keyboard-Friendly-Virtualization.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. Navigating from one child component of a rendered item to another
+1. Arrowing between focused items
+1. Shifting to first/last focused item via home/end keys
+
+`VirtualizedList` has enough internal knowledge to transparently support these
+scenarios. This can be accomplished by adding three **"realization windows"**:
+1. **"focused"** Realize cells around a cell with active focus. Can be
+tracked by listening to bubbling onFocus in the default `CellRendererComponent`.
+1. **"home"** The area at the beginning of the list (for home key).
```
NickGerleman, line 68: Need to noodle a bit more on whether realization of beginning/end should be default or not on desktop. Could imagine it being wasted memory in a lot of circumstances.

NickGerleman:
This has some intersection with existing functionality, where users can mark `stickyIndices` for sticky headers which are kept realized. This solely keeps sticky headers realized though, without additional cells around them.

proposals/0006-Keyboard-Friendly-Virtualization.md
```diff
@@ -0,0 +1,179 @@
...
+
+1. Navigating from one child component of a rendered item to another
+1. Arrowing between focused items
+1. Shifting focus from currently focused item via page up/down
+1. Shifting focus to first/last item via home/end keys
+
+`VirtualizedList` has enough internal knowledge to transparently support these
+scenarios. This can be accomplished by adding three **"realization windows"**:
+1. **"focused"** Realize cells around a cell with active focus. Can be
+tracked by listening to bubbling onFocus in the default `CellRendererComponent`.
```
NickGerleman, line 63: Maybe it would make sense to listen to capture phase instead of bubbling phase, in case something inside the cell wants to stop upward propagation.

appden: I agree this makes sense to track in the capture phase.

NickGerleman:
Pushed out a new commit with a few updates:

- Focus tracking via capture phase
- Smaller realization windows
- Focused -> "most recent focused"
- Talk about home intersection with initial render
- Talk about batch rendering behavior

proposals/0006-Keyboard-Friendly-Virtualization.md
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * Keeps the beginning area of the list realized. Useful to allow instantly
+   * shifting focus to the first item (e.g. via Home key).
+   * - Disabled by default on all platforms (Should be enabled by default on
+   * desktop if FlatList begins transparently supporting home/end focus)
+   *
+   * Defaults to keeping a single screen at the beginning of the list realized
+   */
+  home?: RealizationWindow;
```
appden, line 156: I think we should call this `start` instead. The "home" terminology doesn't really fit well with our other APIs, and "start" is clear.

proposals/0006-Keyboard-Friendly-Virtualization.md
```diff
@@ -0,0 +1,190 @@
...
+### API reference
+
+```tsx
+/**
+ * Allows enabling or disabling an area for realization, optionally overriding
+ * its window size.
+ */
+export type RealizationWindow =
+  | boolean
+  | [boolean, {windowSize?: number}];
```
appden, line 127: I feel like this might be too much unnecessary configuration being able to adjust each `windowSize`. Any reason we'd ever want this to not be 1?

proposals/0006-Keyboard-Friendly-Virtualization.md
```diff
@@ -0,0 +1,190 @@
...
+- **Keep FlatList focus-safe on desktop:** Realization windows needed for
+correct keyboard navigation should be enabled by default.
+- **Keep FlatList memory usage steady on mobile:** New realization windows
+are opt-in where not pay to play.
+
+The full matrix of behaviors is described in the *API Reference* below.
+
+Endpoint detection is generally performed via `Platform` APIs, such as
+`Platform.isTV()` or `Platform.isPad()`. This change would add a
+`Platform.isDesktop()` endpoint query to allow desktop-specific behavior.
```
appden, line 107: Love this! I [suggest this](https://github.com/facebook/react-native/pull/29937#discussion_r491077211) in facebook/react-native#29937 but that never was completed.

proposals/0006-Keyboard-Friendly-Virtualization.md
NickGerleman: The default `windowSize` for the viewport is fairly large, rendering a whole 10 screens above and below. This is a pretty aggressive amount of pre-rendering. It seems like existing setup optimizes for avoiding blanking on fast movement/interaction over memory consumption.

The realization areas have the potential for similar issues. A conceivable user scenario is to quickly tab between items taller than a viewport. Items would possibly not be realized in time for the keyboard interaction. Keeping a larger window fixes this, but a viewport's worth of native controls could be a non-trivial amount of memory, or cause jank due to excessive native renders.

The right amount of prerendering here seems to depend on scenario. I think we can pick something sane as a default, but am not confident a specific size would work well in all scenarios

I think there might be some alternate APIs to make configuration cleaner, and want to play around with that. I think configurability would be useful here in a general-sense though.

proposals/0006-Keyboard-Friendly-Virtualization.md
NickGerleman: I think Android also has this, with`PlatformConstants.constants.uiMode` being 'desk' for desktop.

rozele:
Perhaps somewhat related, I was considering if it might be possible to reduce the number of onLayout events emitted made by items in a FlatList, especially in inverted scenarios where content is refreshed from the to.

From what I could tell, there are at least two uses for onLayout events that aren't in the fixed start/end realization windows. One is for calculating the height of the spacers and the other is for implementing scrollToIndex.

If the native platform offered something to scroll a given component into view, e.g., BringIntoView on XAML or scrollIntoView on Web, we would only need to capture the size metrics of the virtualization candidates, which are significantly less likely to change (but we'd need to add some onSizeChanged or similar event to complement onLayout that could fire less frequently).

In order for something like native scrollToIndex to work, we'd need the ability to realize a window around that view prior to invoking a command, so I'm curious if what you have proposed here would be compatible with this theoretical optimization 😀

NickGerleman:
PRs for internal refactoring of VirtualizedList state representation are now out. This should let us start realizing discontiguous regions.

BasixKOR:
[View Rendered RFC](https://github.com/NickGerleman/discussions-and-proposals/blob/patch-2/proposals/0540-vision-for-layout-conformance-parity.md)

PR #410, opened 2021-09-27
RFC: Accessibility APIs

necolas:
This is a proposal to expand React Native accessibility APIs, and to align those APIs with equivalent web standards, e.g., [WAI-ARIA](https://www.w3.org/TR/wai-aria-1.2/). The aim is to provide developers with a single, familiar accessibility model rather than requiring deep knowledge of the models of each target platform.

### DIFF
--- a/proposals/0000-accessibility-apis.md
+++ b/proposals/0000-accessibility-apis.md
@@ -1,0 +1,156 @@
+---
+title: Accessibility APIs
+author:
+- Nicolas Gallagher
+date: 2021-09-20
+---
+
+# RFC: Accessibility APIs
+
+## Summary
+
+This is a proposal to expand React Native accessibility APIs, and to align those APIs with equivalent web standards, e.g., [WAI-ARIA](https://www.w3.org/TR/wai-aria-1.2/). The aim is to provide developers with a single, familiar accessibility model rather than requiring deep knowledge of the models of each target platform.
+
+## Motivation
+
+To provide React Native developers with a single accessibility model that allows for targeting a wide range of platforms, rather than requiring deep knowledge of the models of each target platform.
+To continue betting on React Native's web origins, as most React Native users have a web background and many of the existing APIs are familiar to web engineers. To avoid designing competing UI standards, and instead leverage existing specs as far as possible, e.g., WAI-ARIA.
+
+This proposal also meets the motivations for previous accessibility proposals related to [lists](https://github.com/react-native-community/discussions-and-proposals/blob/ef368d499b9350c8bcfbfdc8db23b8592982f123/proposals/0000-accessibilityapis-lists.md) and [relationships](https://github.com/react-native-community/discussions-and-proposals/blob/7f6fe5fdba15e4ecd147b69110e81aaf0719e5c4/proposals/0000-a11y-cross-platform.md).
+
+## Detailed design
+
+The web has well-defined accessibility requirements to meet the needs of people using the broad range internet enabled devices and input hardware. Supporting these standards should in turn [benefit native](https://reactnative.dev/blog/2021/08/26/many-platform-vision#embracing-platform-constraints) app accessibility. The accessibility paradigms across Android, iOS, Web, Windows, and macOS have similarities but are not entirely equivalent. A significant challenge lies in shielding users of React Native's accessibility APIs from the expertise otherwise needed to target multiple platforms at once. The rest of this section focuses on specific accessibility concerns and outlines proposed APIs.
+
+### Accessible focus
+
+React Native needs a clear mechanism for marking up elements as focusable by a user action or assistive technology. The `focusable` prop in React Native is currently an Android-only prop to markup elements as *keyboard* focusable. React Native for Web and React Ntive for Desktop use `focusable` to markup elements as *user* focusable, whether by keyboard focus (i.e., appearing in the tab-flow) or as a result of pointer interaction. Elements marked as `focusable={false}` should also provide semantic information, e.g., hiding natively focusable elements from the keyboard tab-flow.
+
+### Accessible lists
+
+React Native needs a mechanism for marking up elements as lists and list items, along with the ability to describe the size of lists and the position of a list item in the set. React Native for Web supports `accessibilitySetSize` and `accessibilityPosInSet` for this purpose.
+
+### Accessible state
+
+React Native currently includes the [`accessibilityState`](https://reactnative.dev/docs/accessibility#accessibilitystate) and [`accessibilityValue`](https://reactnative.dev/docs/accessibility#accessibilityvalue) props to work with limited states and value ranges respectively. However, both of these props are objects. This makes it harder to implement hooks that manage specific accessibility state or value props, as it often requires careful merging of fields that may be managed by different functions. Using objects to manage a mix of static and dynamic properties also has worse perfomance characteristics as updating the value of `accessibilityValue.now` means recreating the entire `accessibilityValue` object, even though the other properties are static. Additional states (e.g., pressed) are not represented in the React Native `accessibilityState` prop. React Native for Web supports all the state and value fields as top-level, primitive-type props, e.g., `accessibilityValueNow`.
+
+### Accessible relationships
+
+React Native would benefit from a way to create accessibility relationships between elements, linked by `nativeID`. React Native for Web has included [`accessibilityLabelledBy`](https://www.w3.org/TR/wai-aria-1.2/#aria-labelledby) and [`accessibilityDescribedBy`](https://www.w3.org/TR/wai-aria-1.2/#aria-describedby) props, which accept an ID or an array of IDs. These props are used in conjunction with `nativeID` to define accessibility relationships between (multiple) elements.
+
+### Phase 1
+
+Renaming existing APIs.
+
+**Example diffs**. Renaming `autoCompleteType` to `autoComplete`: [8066bc9ff](https://github.com/facebook/react-native/commit/8066bc9ff6a3f1d6910ccd346ce0fe6b7b4b16fc), [27fec9569](https://github.com/facebook/react-native/commit/27fec9569e08a04e0dbdbd5de063a599ad0416fa).
+
+* Add `accessibilityBusy: boolean`. Replaces `accessibilityState.busy`.
+    * Add `accessibilityBusy` prop with docs. Spec. 
+    * Deprecate `accessibilityState.busy`.
+* Add `accessibilityChecked: boolean | 'mixed'`. Replaces `accessibilityState.checked`.
+    * Add `accessibilityChecked` prop with docs. Spec.
+    * Deprecate `accessibilityState.checked`.
+* Add `accessibilityDisabled: boolean`. Replaces `accessibilityState.disabled`.
+    * Add `accessibilityDisabled` prop with docs. Spec. 
+    * Deprecate `accessibilityState.disabled`.
+* Add `accessibilityExpanded: boolean`. Replaces `accessibilityState.expanded`.
+    * Add `accessibilityExpanded` prop with docs. Spec. 
+    * Deprecate `accessibilityState.expanded`.
+* Add `accessibilitySelected: boolean`. Replaces `accessibilityState.selected`.
+    * Add `accessibilitySelected` prop with docs. Spec. 
+    * Deprecate `accessibilityState.selected`.
+* Add `accessibilityValueMin: number`. Replaces `accessibilityValue.min`.
+    * Add `accessibilityValueMin` prop with docs. Spec. 
+    * Deprecate `accessibilityValue.min`.
+* Add `accessibilityValueMax: number`. Replaces `accessibilityValue.max`.
+    * Add `accessibilityValueMax` prop with docs. Spec. 
+    * Deprecate `accessibilityValue.max`.
+* Add `accessibilityValueNow: number`. Replaces `accessibilityValue.now`.
+    * Add `accessibilityValueNow` prop with docs. Spec. 
+    * Deprecate `accessibilityValue.now`.
+* Add `accessibilityValueText: string`. Replaces `accessibilityValue.text`.
+    * Add `accessibilityValueText` prop with docs. Spec. 
+    * Deprecate `accessibilityValue.text`.
+* Add `accessibilityHidden: boolean`. Replaces `accessibilityElementsHidden` (iOS) / `importantForAccessibility` (Android).
+    * Add `accessibilityHidden` prop with docs. Spec. 
+    * Deprecate `accessibilityElementsHidden` and `importantForAccessibility`.
+* Add `accessibilityLiveRegion` value of `"off"`. Replaces `"none"`.
+    * Add `"off"` value.
+    * Deprecate `"none"` value.
+* Add `accessibilityModal: boolean`.  Replaces `accessibilityViewIsModal` (iOS).
+    * Add `accessibilityModal` prop with docs. Spec. 
+    * Deprecate `accessibilityViewIsModal`.
+    * Add Android support.
+* Add `accessibilityRole` value of `"slider"`. Replaces `"adjustable"`.
+    * Add `"slider"` value.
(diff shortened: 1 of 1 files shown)


AgneLukoseviciute:
Have you thought about adding [landmarks](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/index.html) to this proposal? We're currently [adding in](https://github.com/microsoft/react-native-windows/pull/8708) support for this in RNW as it's quite important in desktop, might be something that could benefit mobile users as well.

necolas:
Landmark roles are already accounted for in ARIA by using the `accessibilityRole` prop. Additional ARIA roles to add to the RN API are included in this proposal. What where the reasons for deciding to introduce a separate prop for this in RN for Windows? The general idea here is to follow ARIA rather than introduce more proprietary props.

FalseLobster:
Generally speaking I'm very much in favor of this idea!  ARIA seems naturally suited to react-native for a number of reasons:
- It's not browser specific, if you read the WAI-ARIA spec they're very careful to use words like "user agent" to keep it agnostic of its primary use case today (accessibility in web browsers).
- The [CORE-AAM](https://www.w3.org/TR/core-aam-1.2/) does a great job of describing how ARIA attributes should map to native accessibility APIs, since it already is an x-plat abstraction for native accessibility APIs
- It's declarative, which suits react-native well
- The maintainers of the core-aam include representatives from Apple and Microsoft, who have a vested interest in ensuring ARIA can provide a good experience on their platforms.  I think Google may be involved in the draft for the next version.

I think snapping to aria does raise a few questions, though:
- Does this proposal suggest react-native become a _rigorous_ ARIA user agent, or just use APIs that similarly named?  For example, the way [accessible name and descriptions are computed](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te) is fairly complex and may involve some non-trivial native code on the various platforms code (out of tree and otherwise).
- Do you plan to deprecate accessibilityActions as part of the proposal? I didn't see it in the RFC.  A number of native accessibility APIs are imperative in nature and accessibilityActions is a fairly nice way to add that functionality.  It's not really possible to write a fully compliant Slider control on Windows using just ARIA because in ARIA all of the interaction is through keyboarding and simulated DOM clicks.

necolas:
> Does this proposal suggest react-native become a rigorous ARIA user agent, or just use APIs that similarly named?

Being pragmatic it's the latter for now. Although I think having specs for implementations is part of the appeal (as it is for Yoga to implement flexbox), and at worst the specs can be used to understand how the RN implementation differs from others.

> Do you plan to deprecate accessibilityActions as part of the proposal?

That is the open question at the end of the proposal. In the short term these props are needed. If other changes to RN are made in the future we might be able to separate behavior from views, including platform-specific functionality like accessibility actions.

```jsx
const ref = useAccessibilityActions({
  onAccessibilityAction,
  accessibilityActions
});

return <View ref={ref} />;
```

Problems like this lead me to ask: 1) Can `accessibilityActions` be inferred from ARIA/Web patterns? or 2) Would the Web benefit from APIs equivalent to the imperative native APIs accessed via `accessibilityActions`?

AgneLukoseviciute:
> Landmark roles are already accounted for in ARIA as values of the `role` attribute (equivalent to `accessibilityRole` prop). The additional ARIA roles (inc landmarks) to add to the RN API are [included in this proposal](https://github.com/react-native-community/discussions-and-proposals/blob/fd64774e8161973b87b8f0a27587fa039eb7b43b/proposals/0000-accessibility-apis.md?plain=1#L113-L115) and already supported by RN for Web. What where the reasons for deciding to introduce a separate prop for this in RN for Windows? The general idea here is to try to follow ARIA as closely as possible to avoid introducing proprietary props.

I hadn't realized landmarks in ARIA are implemented as values for role, thanks for pointing it out. That's the main reason I went with an implementation through a new prop, but we'll be able to add these values and move the logic over to accessibilityRole fairly easily.

ksiler:
I'm also in favor of this proposal 🙂 my questions are:

- Have you considered adding properties `accessibilityMultiSelectable` and `accessibilityRequired` that would be equivalent to [aria-multiselectable](https://www.w3.org/TR/wai-aria-1.1/#aria-multiselectable) and [aria-required](https://www.w3.org/TR/wai-aria-1.1/#aria-required)? We currently support these concepts in react-native-win32 as part of the accesibilityState property and it's helpful for our native APIs when dealing with a group or list of elements that can be selected.

- What's the reasoning for removing [`accessibiltyHint`](https://reactnative.dev/docs/accessibility#accessibilityhint) in favor of just using [`accessibiltyLabel`](https://reactnative.dev/docs/accessibility#accessibilitylabel)? I thought these properties had distinct purposes. How should developers convey extra information about a control if `accessibilityHint` is removed?

- To clarify, is the proposed [`accessibilityLevel`](https://github.com/react-native-community/discussions-and-proposals/blob/fd64774e8161973b87b8f0a27587fa039eb7b43b/proposals/0000-accessibility-apis.md?plain=1#L109-L110) property meant to refer to both text heading levels and levels of elements in a hierarchical structure (which can be inferred by the role)? I'm asking because we already have an `accessibilityLevel` property for react-native-win32 but it's just used for the latter scenario in conjunction with `accessiblityPosInSet` and `accessibiltySetSize`. We could easily repurposed it for identifying text heading levels too but I wanted to check if both scenarios fit the intended use.

necolas:
> Have you considered adding properties `accessibilityMultiSelectable` and `accessibilityRequired` that would be equivalent to `aria-multiselectable` and `aria-required`? We currently support these concepts in react-native-win32 as part of the `accessibilityState` property and it's helpful for our native APIs when dealing with a group or list of elements that can be selected.

Good to know you already need these props too. All the ARIA properties are [supported in RN for Web](https://necolas.github.io/react-native-web/docs/accessibility/) and if we were to adopt the ARIA APIs they could be added to RN. I can include them in the proposal. There are even more ARIA attributes for defining tabular data, etc...do you also have a need for those yet?

> What's the reasoning for removing `accessibilityHint` in favor of just using `accessibilityLabel`? I thought these properties had distinct purposes. How should developers convey extra information about a control if `accessibilityHint` is removed?

The `accessibilityHint` prop is an iOS-only concept. There is an attempt to implement it in RN for Android, but it's hacky and even introduces accessibility issues related to localization. There doesn't appear to be a clear need for this prop as the label should confer the necessary information and exists across all major platforms. Furthermore, the "additional information" use case is served by `aria-describedby`.

(This proposal is an attempt to provide guiding principles - ARIA - for defining new a11y props so we can limit further proliferation of adhoc prop additions.)

> ...is the proposed `accessibilityLevel` property meant to refer to both text heading levels and levels of elements in a hierarchical structure

Yes I think we should support both since that is the attribute's purpose in ARIA too. But it depends on the constraints of various native platforms and requires input from Android/iOS experts.

blavalla:
While I like the general idea of having a more unified API between native and web, there are some pretty big differences that would need to be taken into account that would ultimately make this new API not quite close enough to web to be a direct mapping, but not quite close enough to native to also be a direct mapping, basically ending up as a strange hybrid of the two (which doesn't even take into account the large differences between how iOS and Android handle some of the basics like focus).

To give some context, web accessibility has a very declarative API, where you declare various attributes on DOM elements, and then allow the assistive tech (AT) to parse those out and determine its behavior. On native, while many properties are declarative (traits, labels, etc.), there is also the concept of directly interfacing with the AT itself to control it. Things such as sending accessibility events on Android, creating custom rotors on iOS, and making manual announcements (both platforms), all rely on the app directly telling the AT to do something, and it responding to this.

This difference is likely due to the significantly simpler interaction models that mobile ATs have, where a user may only have a few gestures they can use to control them, so much of the advanced control relies on the apps themselves to add specific actions and features for the user.

As for the details of this proposal, I have a few items that I think need to be thought through in more depth. I'll outline them by phase below:

Phase 1
- Consolidating accessibilityElementsHidden and importantForAccessibility works for common cases, but not more advanced ones.  On Android for example you may want to set importantForAccessibility to "no_hide_descendants", or "no", one of which blocks this element and *all* descendants from this node onwards from being accessible, and the other one only blocks this specific element.  If this became a boolean this functionality would be lost.  This also disallows the setting of this value to "yes", although admittedly that has far fewer use cases (though not zero).
- Replacing the "adjustable" role with "slider" doesn't make much sense from an iOS perspective.  Many non-slider elements are considered "adjustable", and this role maps directly to the UIAccessibilityTraitAdjustable trait, which is used for a number of interfaces where a user adjusts a value. In general, there is a real problem right now with the combining of "roles" (from web and Android) and "traits" (from iOS).  Roles only describe what an element *is*, while traits are used to both describe what an element *is* and also how an element *works*.  Trying to use one prop for both of these concepts means that there will always be strange edge cases like this.
- accessibilityHint is a very important feature on mobile, and can not be replicated by simply concatenating content into the label. Hints are purposely delayed after the main focus announcement, as they are meant to describe the result of taking an action on the element. A user should not need to hear them every time they land on an element, as they would become annoying quick quickly in UIs that a user understands. They are meant to teach the user how the UI works, and be easily ignored (hence the delay) if they are no longer needed. Android *does* have a concept of hint, but on Android hints are directly connected to accessibility actions, so a hint can't be a standalone string like on iOS, but must be associated with an action such as "click".  We surface these in RN right now via the `accessibilityActions` API due to this required association with an action.

Phase 2
- I think we'll need to flesh out what these props actually *do* on mobile, as many of them either have similar equivalents already (posInSet compared to accessibilityValue min/max/now) or have no mobile counterpart (describedBy, activeDescendant, etc.).  
- The biggest issue with the Phase 2 items is the concept of redefining "focusable" semantics.  Android can map very closely to web here, but iOS simply does not allow a focusable element to have any focusable descendants. There is no way around this without basically re-creating an entirely custom hierarchy of accessibilityElements (that is still in a flat list, but ordered in a way that makes it *seem* like you are navigating to descendants) that is not connected to the actual view hierarchy.  

Phase 3
- My only feedback here is that you've removed accessibilityIgnoresInvertColors without any sort of way to get this feature back via some other API. So this would just be a feature loss, that while not the most commonly used feature, has no other viable approach to fall back on if you do need it. The other changes all have had alternate APIs suggested, so their removal is okay assuming that those replacement APIs are well fleshed out.

So in conclusion, I really like the idea here, but I think it likely needs to have someone with expertise on each platform (web, Android, iOS, Windows) to really be involved to make sure we don't miss anything or leave any weird gaps. It's certainly possible to make a more unified API, but that API may not map as closely to one platform (web) as initially thought.

Let me know if you want details on any of the above. I would consider myself an expert in Android accessibility that can speak to any part of that system, and knowledgeable enough about iOS accessibility to make high level suggestions, but not enough of an expert on iOS to give in depth explanations of the technical details required to pull something off.

necolas:
Thanks Brett. Are there any other native a11y APIs that are currently unavailable in RN?

> On native, while many properties are declarative (traits, labels, etc.), there is also the concept of directly interfacing with the AT itself to control it. Things such as sending accessibility events on Android, creating custom rotors on iOS, and making manual announcements (both platforms), all rely on the app directly telling the AT to do something, and it responding to this.

If developers need to understand and manage 3+ different a11y APIs for general a11y tasks while using React Native, that would be a serious problem. What we have to decide is a) whether the differences are significant enough that the imperative controls cannot be abstracted behind the ARIA declarative API, and b) whether the use cases are common enough that they must belong in the xplat API (but only be supported by 1 or 2 platforms). There will always be a need for platform-specific APIs in specific scenarios, but I think we should avoid putting those APIs into the xplat layer. We have other mechanisms, such as custom components and hooks, by which platform-specific behaviors could be introduced.

> On Android for example you may want to set importantForAccessibility to "no_hide_descendants", or "no", one of which blocks this element and all descendants from this node onwards from being accessible, and the other one only blocks this specific element. 

Is this not mostly equivalent to `aria-hidden` (hides tree) and `role="none"` (removes semantics from individual element)? See [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute).

> Replacing the "adjustable" role with "slider" doesn't make much sense from an iOS perspective.

The `slider` role isn't intended only for visual sliders, but for UI where a value is selected from within a range. There may well be differences between traits and roles, but RN has already combined the 2 concepts.

> accessibilityHint is a very important feature on mobile, and can not be replicated by simply concatenating content into the label...We surface these in RN [for Android] right now via the accessibilityActions API due to this required association with an action.

Concatenating the content into the label is exactly what we do for the Android implementation of `accessibilityHint`, and since it's not done in user space this is a big problem for localized content. `aria-describedby` appears to be the rough equivalent of hints, in terms of the purpose and precedence it takes. If we were to keep `accessibilityHint` it would be as an iOS-only prop. Longer-term I don't think the `accessibilityActions` API belongs on the component props, but as a behavior hook. This proposal doesn't suggest removing that API at this stage though.

> I think we'll need to flesh out what these props actually do on mobile

Agreed. If we were to do this it looks like we'd be committing to building certain abstractions over native APIs as browsers do. I have no idea what that involves on the native engineering side.

> iOS simply does not allow a focusable element to have any focusable descendants

Technically neither does web. This is something we should enforce at the framework level in any case.

> you've removed accessibilityIgnoresInvertColors without any sort of way to get this feature back via some other API

This is another iOS-only prop that has no relation to ARIA concepts. I'm not sure what the best replacement would be. I think we could either make iOS behave like other platforms at the framework level (maybe that means media is never inverted?), or perhaps model the API to control this on web equivalents (e.g., the [forced-color-adjust](https://www.w3.org/TR/css-color-adjust-1/#forced-color-adjust-prop) style).

necolas:
RE: Accessibility Actions

This Apple introduction to custom rotors is interesting: https://developer.apple.com/videos/play/wwdc2020/10116/

The first example of the categorized map icons would typically be implemented on web as buttons in either a list of lists, or lists below headings. VoiceOver users can then use the rotor to navigate by category headings. Other examples could perhaps be implemented by React Native, mapping web markup patterns to inferred custom rotors on native platforms. While the rest don't have clear equivalents.

There's been some discussion in the W3C ARIA GitHub about how to model these accessibility actions, suggesting they have similarities to custom context menus for accessibility services: https://github.com/w3c/aria/issues/762

nonoomir (approved):

ksiler:
@necolas We're currently looking at adding support for defining text heading levels in [react-native-win32](https://github.com/microsoft/react-native-windows/tree/main/packages/%40office-iss/react-native-win32#office-issreact-native-win32) by reusing the existing [`accessibilityLevel`](https://github.com/microsoft/react-native-windows/blob/main/packages/%40office-iss/react-native-win32/src/Libraries/Components/View/ViewWin32.Props.ts#L215) prop we have defined which is also referenced in the proposal [here](https://github.com/react-native-community/discussions-and-proposals/pull/410/files#diff-0778a9f2187181434c1b3042681941d44513d93af9c5dcbbd1b3e94e329c140cR109). Since I last asked about this over a year ago, I wanted to reconnect and verify we could move forward with this. Here is our previous conversation:

>> ...is the proposed accessibilityLevel property meant to refer to both text heading levels and levels of elements in a hierarchical structure

>Yes I think we should support both since that is the attribute's purpose in ARIA too. But it depends on the constraints of various native platforms and requires input from Android/iOS experts.

Are there any android/iOS contacts I should reach out to to discuss this? Or is the proposal outlined for [`accessibilityLevel` ](https://github.com/react-native-community/discussions-and-proposals/pull/410/files#diff-0778a9f2187181434c1b3042681941d44513d93af9c5dcbbd1b3e94e329c140cR109) approved?

blavalla:
> 

Just as something worth noting, I don't think there is any Android equivalent to heading levels, so this property likely will be a no-op on that platform. The only place that Android has any sort of level information like this is with AccessibilityCollectionInfo, which has a boolean isHierarchical saying whether this list should be treated as hierarchical content and whether nested lists should be considered as different levels.

There's no control over the actual level other than the depth of the list though, so even if this was set on elements within hierarchical list structures it wouldn't really be applicable.

necolas:
Closing this in favor of #496

PR #480, opened 2022-05-19
RFC: React Native as a monorepo

kelset:
This proposal is something that me and @tido64 (and @cortinico to some extent) have been brainstorming for a little while - after a recent problem with `react-native-codegen` and RN68.

As per title, basically this proposal is not much as in introducing something completely new, but more to discuss a solution to an existing problem. To put simply, the current shape of the `react-native` codebase hosted in GitHub is lacking structure and consistency and this is leading to a number of issues.

You can read more details in the text itself - here's the [file view](https://github.com/kelset/discussions-and-proposals/blob/kelset/proposal-react-native-monorepo/proposals/0006-react-native-monorepo.md) for easier reading experience.

<details>
<summary>To see proposal in graph form, expand this.</summary>

RN as a monorepo proposal

</details>

Feel free to add comments and questions! I'd love to see this turn into something actionable as soon as possible, since it addresses a number of system problems with the repo - it will require some coordination and collaboration, but I'm sure Meta and MSFT can collaborate to make it happen :)

This will also unlock/make easier to revisit other RFCs, such as @empyrical's https://github.com/react-native-community/discussions-and-proposals/pull/49

---

*sidenote: this pr also adds a folder to keep images in, `assets/`, and adds to the .gitignore the pesky macos DS_Store*

---

last updated on Sept 2022 to reflect the latest status.


proposals/0006-react-native-monorepo.md
```diff
@@ -0,0 +1,166 @@
...
+  - in particular, we recommend adding the tool [`changeset`](https://github.com/changesets/changesets) to the codebase to handle the release coordination (as in, once 0.70 gets created, all the packages in that branch get versioned to `0.70.x`) and bump packages accordingly when new versions are needed
+
+_sidenote_: we recommend this work to be done all in the same timeframe between a minor branch cut and the next (ex. `0.70-stable` gets created, then this works start, and until all of it has been done, `0.71-stable` is **not** created).
+
+In closing this section, we also want to acknowledge how this proposal is deliberately not introducing any high degree of automation or advanced tooling - this is because we are well aware that this repository is but a "partial mirror" of how the react-native code is shaped within the Facebook monorepo, and adding more extensive and invasive tooling would require also introducing them to that monorepo. So we opted for the minimal footprint that would be OSS-side only (with the tradeoff of more custom, local code and scripts).
```
cortinico, line 113:
```suggestion
In closing this section, we also want to acknowledge how this proposal is deliberately not introducing any high degree of automation or advanced tooling - this is because we are well aware that this repository is but a "partial mirror" of how the react-native code is shaped within the Meta monorepo, and adding more extensive and invasive tooling would require also introducing them to that monorepo. So we opted for the minimal footprint that would be OSS-side only (with the tradeoff of more custom, local code and scripts).
```

proposals/0006-react-native-monorepo.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Other considerations
+
+- we could consired renaming `react-native` into `@react-native/core` but no strong opinion; given how this is the "front-facing" package it might be easier and more digestable for consumers of react-native to just keep stay on the current convention.
+
+## Drawbacks
+
+- Most of it needs to be tackled in "one fell swoop"
+- Old open pull requests will likely need to be rebased and re-arranged
+- Some of the native build scripts (gradle, BUCK, xcode, etc.) and CI scripts and config will need to be updated for the new code locations
```
cortinico, line 138:
```suggestion
- Some of the native build scripts (Gradle, BUCK, XCode, etc.) and CI scripts and config will need to be updated for the new code locations
```

cortinico (commented):
Thanks for sending this over @kelset 
I've did a first pass and left some comments/nits

proposals/0006-react-native-monorepo.md
```diff
@@ -0,0 +1,166 @@
...
+This proposal will seem deceptively simple: to decouple the root level of the package from being both root and `react-native`, to just be the former, and move `react-native` as a package in the `packages/` folder.
+
+The graph here shows in a bit more detail the end goal, along with a few more details and side-changes that are needed to fully address the current situation:
+
+![proposal for the react-native GitHub codebase](/proposals/assets/0006-proposal.png "Proposal")
+
+Reiterating the changes proposed in the graph:
+
+- step 1 is to move the `react-native` specific code in its own folder within the `packages/` folder. This is the most error prone step to handle, as many paths and variables might have to be modified (for ex. in the CI configurations) to accommodate for this change.
+- step 2 is to **rename** some of other packages as shown above, and in this table:
```
yungsters, line 91: Would we want to also standardize naming of the `react-native/packages/*` subfolders at the same time?

mikehardy:
Just as a supporting comment, my experience supporting the firebase community leads me to lean strongly to "locked versioning" for easy of user support / issue triage as it eliminates much of the versioning confusion - firebase ecosystem has transitioned fully to that style (if you'll allow that a gradle "bill of materials" fits the bill on android - since we use the BoM - though their components do have independent versioning underneath). It really helps a lot, so having "one true version" with automated release policy would be nice from my perspective. (https://invertase.io/blog/react-native-firebase-versioning)

If some packages are higher velocity than others leading to faster versions and worries about release eng time, my counters are "more automation" and "versions are free". I recognize "more automation" isn't free, but within reason it should be possible, if faster revving packages are an issue.

Not sure if there are other counterarguments but may well be, happy to learn if so, especially if they really are insurmountable and mean locked versioning is a non-starter in some cases

proposals/0006-react-native-monorepo.md
```diff
@@ -0,0 +1,166 @@
...
+  - [Implementation detail considerations](#implementation-detail-considerations)
+  - [Other considerations](#other-considerations)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+- [Future work](#future-work)
+- [Unresolved questions](#unresolved-questions)
+
+## Motivation
```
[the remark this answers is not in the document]
cipolleschi: I would also add that his setup makes hard to have consistent testing in the CI:
* we have two different build patterns for RNTester and for the templates.
* templates has to do magic trick to simulate an npm package for `react-native` when testing.

proposals/0006-react-native-monorepo.md
cipolleschi: [nit] The Apple's IDE proper name is **Xcode** 😉

cortinico: Updated 👍

proposals/0006-react-native-monorepo.md
kelset: added a separate table with subfolders name migration, and updated the proposal graph to match 👍

proposals/0006-react-native-monorepo.md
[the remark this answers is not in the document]
kelset: thanks for your inputs! I've added them to the section :)

proposals/0006-react-native-monorepo.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Implementation detail considerations
+
+- This work will have to start from Meta's side (see [comments in the other proposal](https://github.com/react-native-community/discussions-and-proposals/pull/49/files#r255135557)): there's a lot of company specific knowledge involved in how the `react-native` repo gets created for GitHub, and in how the code is shaped in the internal monorepo.
+- Since Metro by default doesn't handle symlinks well (see [this issue for details](https://github.com/facebook/metro/issues/1)), it could be that for RN Tester to keep working we might have to add [`@rnx-kit/metro-resolver-symlinks`](https://github.com/microsoft/rnx-kit/tree/main/packages/metro-resolver-symlinks)
```
motiz88, line 156: `@rnx-kit/metro-resolver-symlinks` is not vetted for correctness and its use can lead to bugs, as it sidesteps Metro's existing filesystem abstractions and reads directly from disk. Let's not bring it into the RN repo and instead look into what's required to get the new structure working when we get there (e.g. possibly a more constrained & safe custom resolver).

kelset: makes sense - that's why I kept it conditional ;)
that aside, let me cc @tido64 and @afoxman so that they are aware of these feedbacks about our symlinks solution; maybe (if and when we need a solution) we can figure out how to make it more compliant

empyrical:
One approach I was thinking about for this initial monorepo restructuring was a codemod script of some sort - so you don't have to keep rebasing a huge pull request keeping up with the main branch, you just need to maintain a script which restructures the repo/applies needed patches until it's decided the changeover is ready.

There's a lot of internal stuff at Meta that would be effected by changing the structure of the RN repo too so this effort will need involvement from someone there. A codemod script approach would probably also be useful because once it's to the point where tests are passing on OSS with the new structure, someone at Meta could modify/add on to the script to change over things internally too.

And hopefully it could be useful for helping people rebase their pull requests to the new structure when the change happens, too.

cortinico:
> One approach I was thinking about for this initial monorepo restructuring was a codemod script of some sort - so you don't have to keep rebasing a huge pull request keeping up with the main branch, you just need to maintain a script which restructures the repo/applies needed patches until it's decided the changeover is ready.

That's a good point. I still think that the major challenge at this stage is letting this setup works with the internal infrastructure. Having one of your draft PR @empyrical  imported internally as a first run could help us at least draft a list of the necessary pieces to make this happen.

kelset:
> Do we feel the first two items are actionable in isolation? Or do we feel we can split this work further or differently? I just want to make sure this doesn't get stale as it looks too much of a monolithic work.

I think we could start looking at #2 first, put it in place, then do #1.

And yeah #3 can/should be tackled separately

cortinico:
> I think we could start looking at #2 first, put it in place, then do #1.

Yes the publishing (.2) could be done also first. The only gotchas there is make sure we don't re-publish older versions of packages which are already published (as they will fail to do so). Still is something we can split and do separately.

Adding a couple of feedbacks/thoughts on the overall RFC:

- Before we do anything, it would be great to understand how the monorepo is going to be managed. Are we going to use Lerna or any other monorepo mgmt system? If yes, it would be great to have a bit of evaluation of them before we jump into solution mode.

- An open problem which is still unresolved, and is unique of `react-native` is storage of Native Artifacts (Android's `.aar` and Tarballs for iOS) that can't fit inside the NPM package due to size constraints. In general it would be great to find a solution to this problem as well as those artifacts are currently thigthly coupled to NPM packages. I don't want to increase the scope of this RFC, so we should probably move the discussion somewhere else (to another RFC?). Just keep in mind that those artifacts will have to be handled if we move things around.

cortinico:
> How much storage do we get from GitHub? Could we store binary blobs here and download them on demand?

GH Release would be a good solution. The problem is that we would like to have a solution that works also for nightly without having to publish a new Github Release every day as it will create a lot of noise in the repo.

tido64:
> > How much storage do we get from GitHub? Could we store binary blobs here and download them on demand?
> 
> GH Release would be a good solution. The problem is that we would like to have a solution that works also for nightly without having to publish a new Github Release every day as it will create a lot of noise in the repo.

Are you referring to the number of releases or that latest would point a nightly? I imagine this tick box should at least address the latter:
Screenshot 2022-07-20 at 20 50 02

cortinico:
> Are you referring to the number of releases or that latest would point a nightly?

To the number of releases. I'm afraid we could end up in situation like this:
https://github.com/adoptium/temurin18-binaries/releases

Having hundreds of releases to crawl + generating yet another notification for users subscribed to the repo. Maybe I'm overestimating the impact of this and going with Pre-release for nightlies would also work.

kelset:
I agree with Nico here, I am not a fan of having too many GH releases - let's maybe consider an alternative, like publishing them as a new npm package separately? ex. `@react-native/hermes-artifacts`?

cortinico:
> like publishing them as a new npm package separately

The root issue here is sizing. The Hermes tarball is ~400Mb and can't stay in a NPM package. Splitting is 'doable' but would add a lot of custom infra to re-compose the artifacts so really suboptimal

kelset:
So basically it'd be better to have something like a "bucket", sourceforge style thing where to store them. Right. Need to think about this.

tido64:
> The root issue here is sizing. The Hermes tarball is ~400Mb and can't stay in a NPM package. Splitting is 'doable' but would add a lot of custom infra to re-compose the artifacts so really suboptimal

What does that tarball include? [`hermes-engine`](https://www.npmjs.com/package/hermes-engine) is only 20 MBs.

cortinico:
> So basically it'd be better to have something like a "bucket", sourceforge style thing where to store them. Right. Need to think about this.

Essentially. An S3 bucket would work but anything where we can store asset and access them via URL could work. Worth noting that CircleCI assets might already solve this issue. Artifacts have a default retention of ~30 days, which I believe is acceptable for nightlies: https://circleci.com/docs/artifacts#artifacts-and-self-hosted-runner or we can host them somewhere else.

> What does that tarball include? [`hermes-engine`](https://www.npmjs.com/package/hermes-engine) is only 20 MBs.

Debug symbols are making the tarball explode both on Android & iOS. The iOS tarball here is 466Mb: https://github.com/facebook/react-native/releases/tag/v0.69.2
The Android equivalent had to have Prefab libs stripped as they come with Debug symbols by default and would have bumped the NPM package to ~300Mb: https://github.com/facebook/react-native/blob/ccdf9ac9853601a81ff21f8f42f5bd866dd4de75/ReactAndroid/hermes-engine/build.gradle#L52-L54

This is only for Hermes. All the RN libraries are shipped without Debug symbols due to size constrains, making for a far from optimal debugging experience IMHO.

renchap:
I am not familiar with RN's infrastructure, but I think you can take inspiration from the way Homebrew manages their pre-built binaries (Bottles) using Github's package registry. I am not an expert on it, but I remember reading some threads on the topic and that their solution has been advised and validates by Github

They are using Github Packages to host them. Github Packages are not tied to releases or a repo per default (they belong to a user or org), so they wont pollute the Releases page, but they can optionally be linked to a repo.

The trick is to upload the artifact as an OCI image, a container with only one layer. OCI images are a group of tgz archives and the content does not matter.

They use Skopeo for it, like this:
```
/usr/bin/skopeo copy --all oci:artifact--version docker://ghcr.io/homebrew/core/artifact:version
```

From what I understand, the `artifact--version` directory needs to be compliant with [OCI's image layout ](https://github.com/opencontainers/image-spec/blob/main/image-layout.md), but this basically means writing a few JSON files and putting your tarballs at the correct location.

Once uploaded, the artifacts can be downloaded using a stable URL, and discovered using GH package API, without authentication. Github can cache them efficiently as all artifacts have a public and static URL once uploaded.

This is probably a different discussion than this RFC, but I think this would elegantly solve the artefact distribution problem without having to manage external system or credentials. If needed, I can spent some time investigating this further and work with a React Native team member on the implementation.

Some links:
- feature announcement (for external taps): https://brew.sh/2020/11/18/homebrew-tap-with-bottles-uploaded-to-github-releases/
- all the core packages then moved to it: https://github.com/Homebrew/brew/pull/11070
- a sample Github action uploading an artifact: https://github.com/Homebrew/homebrew-core/runs/7698895866?check_suite_focus=true

kelset:
thanks @renchap for the great insight! Worth looking into it more :)

---

Just wanted to add a comment here to quote some things that @yungsters mentioned in his PR (https://github.com/facebook/react-native/pull/34401) so that we won't lose track: 

> * Should I separately configure Flow (and include flow-bin in devDependencies)? ESLint? Jest? TypeScript? Babel?
> * How should I verify that @react-native/event-emitter references in react-native work correctly, without having to publish merge a pull request (which might break things), publish @react-native/event-emitter, and then re-test locally? I think that's what Yarn Workspaces is supposed to be for… but what are the right steps to verify?
> * My intuition is that npm packages should be compiled to run in as many environments as possible, but for something like this… is that really necessary? It's a nice to have, but React Native provides some nice assumptions (e.g. Metro support for ES Modules and Flow). Should I be manually producing X.js that has Flow removed via Babel, but also include X.js.flow? It would be easier if X.js could just be the same file in the react-native repository source.

We should probably enhance the RFC with a section along the lines of "expectation on packages in the /packages folder" or something with some guidelines for when adding a new one or extracting one from react-native itself. This would also help with scenarios like the one recently mentioned in @necolas' RFC https://github.com/react-native-community/discussions-and-proposals/pull/496#discussion_r944773796 of extracting Animated to its own package (that was previously attempted by @EvanBacon here https://github.com/facebook/react-native/pull/28892)

cortinico:
> I am not familiar with RN's infrastructure, but I think you can take inspiration from the way Homebrew manages their pre-built binaries (Bottles) using Github's package registry. I am not an expert on it, but I remember reading some threads on the topic and that their solution has been advised and validates by Github

I haven't followed up on this yet, but I'd like to mention that this is a great solution and might actually be the silver bullet for our problems! I've reached out to @renchap and we'll try to come up with a solution around this.

I'd say for the time being, let's trying to keep the conversation here focused on Monorepo (folder structure, versioning, releasing, etc.). We'll be sharing a separate discussion on packaging/artifacts in the future.

kelset:
Adding an extra note here, we will likely have to do some deep-dives across the react-native repo and figure out if there are rogue packages hidden within it. By complete chance I just found out about this extra package.json living deep within the ReactCommon folder... https://github.com/facebook/react-native/blob/main/ReactCommon/hermes/inspector/tools/msggen/package.json

this type of scenario should not be present anymore when the goal of this RFC is completed (a hidden, unpublished, package.json hidden deep within the source code)

necolas:
> This would also help with scenarios like the one recently mentioned in @necolas' RFC https://github.com/react-native-community/discussions-and-proposals/pull/496#discussion_r944773796 of extracting Animated to its own package (that was previously attempted by @EvanBacon here https://github.com/facebook/react-native/pull/28892)

And it would likely help us avoid Animated becoming more coupled to React Native internals, as has recently happened with the introduction of `ReactNativeFeatureFlags` into the module and the integration of TurboModule calls in a way that assumes native-only use of `Animated`. 3rd party offerings like Reanimated 2 avoid these issues and consider web support along the way.

cortinico:
> We'll be sharing a separate discussion on packaging/artifacts in the future.

Re to @kelset @tido64 @renchap about this. Discussions around artifacts distributions for React Native should go here: https://github.com/react-native-community/discussions-and-proposals/pull/508

kelset:
related, I've set up a dedicated label to track all the work that is going on [Tech: Monorepo](https://github.com/facebook/react-native/issues?q=is%3Aopen+sort%3Aupdated-desc+label%3A%22Tech%3A+Monorepo%22)

cortinico:
Hey all,
I wanted to share an update here as we started doing some practical work on this RFC during the Core Contributor Summit 2022.

After several brainstorming, we envision the work split in those phases:

## Phase 1 (Package Renaming & Reversioning)

Package renaming & versions alignment. We already have some PR opens on this side which is great! I've created an umbrella task to keep track of the work and give the opportunity to others to also jump on this:
https://github.com/facebook/react-native/issues/34692

As a rule of thumb, we decide to **not** do folder renaming at this stage. This is causing a lot of friction as it requires a lot of internal handling on our end, and we believe it brings little value, at least at this early stage.

Ideally we should be done with all the renaming by 0.71.x so that we can annonce that this change has happened in the release notes. We don't expect any user facing impact as users should depend only on `react-native`. Still it's worth to make sure that all the renaming/bumps are bundled together in a single version.

## Phase 2 (Automation)

Once all the versions are aligned, we can work on automation to support:
- Publishing of all the packages from the CI
- Automating the bumping of the version

## Phase 3 (Moving the `react-native` package)

Once we're done with the setup, we can look into moving the top level package inside `packages/react-native` and have a proper yarn workspace in the top level `package.json` (by also removing the `repo-config/` folder).

## Phase 4 (Cleanup)

Once we're done with the full monorepo setup, we'll into follow-up tasks such as:
- Deprecating older packages on NPM (those that have been renamed)
- Remaing folders, for those we believe we'll get value out of the rename.

kelset:
(sidenote: maybe we might want to use this as tool https://github.com/ghiscoding/lerna-lite - it's also used by Jest)

Pranav-yadav:
> This package is still tightly coupled to RN version, so we would need to do something more specific. _--PR specific_
> #### But we also have the issue where the main branch and other branches are unversioned. So whatever does the bumping really needs to be upping the peerDependencies of the monorepo packages to match the new RN version. @cortinico @cipolleschi will the current version bumping script bump monorepo packages peerDependencies on RN?

Qn posted in one of the monorepo pkg related PR review by @NickGerleman; https://github.com/facebook/react-native/pull/35406/files#r1028455787

<sub>Thought, it's worth mentioning it here..</sub>

cortinico:
Hey all,
Just a small heads up that we started merging several of the community PRs we received. Sorry for taking a bit longer but this required a bit of rework for us internally. @hoxyq has been driving this work so far (kudos to him for doing so!). More on this here: https://github.com/facebook/react-native/issues/34692#issuecomment-1319190364

We should be completing Phase 1 in the near future as all the PRs will be merged.

cortinico:
## Update

Phase 1 is now complete, kudos to @hoxyq for driving this forward:
https://github.com/facebook/react-native/issues/34692#issuecomment-1339015561

We'll now start looking into Phase 2 (publishing/bumping automation) and Phase 3 (moving to `packages/react-native`).

kelset:
# Update

Phase 2 is now in good shape, with a few PRs from @hoxyq like https://github.com/facebook/react-native/pull/35767 we should have the main automation all in place in main branch (and so for 0.72+)

hoxyq:
> Should we also update the documentation on the Website?

We have a [wiki page](https://github.com/facebook/react-native/wiki/Release-and-its-automated-processes) for it. As was discussed with @kelset, we will update documentation on the website after we fully validate it on some release (probably next one). In case if we will need to update / change something.

proposals/0006-react-native-monorepo.md
```diff
@@ -0,0 +1,229 @@
...
+- [Adoption strategy](#adoption-strategy)
+- [How we teach this](#how-we-teach-this)
+- [Future work](#future-work)
+- [Unresolved questions](#unresolved-questions)
+
+## Motivation
+
+To quickly explain the motivation behind this proposal, here's the current shape of the React Native codebase hosted on GitHub and corrispective npm packages:
+
+![current status of the react-native GitHub codebase](/proposals/assets/0006-current-status.png "Current Status")
```
huntie, line 39: Can we use a relative URL for both images? Not displaying in GitHub preview and unclear if this will work in other previewers.

kelset: I'm not sure which preview you are referring to, but when I look it here https://github.com/kelset/discussions-and-proposals/blob/kelset/proposal-react-native-monorepo/proposals/0006-react-native-monorepo.md they should up?

Also, I guess you mean absolute urls? ex. https://github.com/kelset/discussions-and-proposals/blob/kelset/proposal-react-native-monorepo/proposals/assets/0006-current-status.png ?

huntie: Fair! Where I saw this not working was in the [PR files preview](https://github.com/react-native-community/discussions-and-proposals/pull/480/files/4ea9fb3e93e3a814812d406551e07dd7d4b3fe4a?short_path=d97c7fa#diff-d97c7fabe0f53f5637e08b32c195b4fc945add1cf98ebeb34ec2279f9e577cbc). I'd expect that `` might work in more places, such as local text editors also. Super minor point.

kelset: ohhh I see, ok no problem let me address that - will tackle it tomorrow

Iamher87 (approved):

hoxyq:
Hey, wanted to share a quick update on this RFC.

I am currently working on making this migration to `react-native` package, you can track progress and leave comments here - https://github.com/facebook/react-native/pull/36270.

We have decided to make it in one go, it seems like a more feasible solution, all previous attempts to setup some infra around to migrate some folders one at each time have had no success. We also don't want to have repository in such transitional state where it can lead to having some weird workarounds for a long time.

Essentially, we will move files which [are included in react-native's npm package](https://github.com/facebook/react-native/blob/main/package.json#L17-L74). There should be no changes for package consumers.

Luckily, we will get rid of `repo-config`, but all `devDependencies` will be listed in root `package.json`, and `/packages/react-native` will list only `dependencies`. This comes from our internal setup (cc @motiz88), and I hope we can improve this later as a separate project.

proposals/0006-react-native-monorepo.md
```diff
@@ -0,0 +1,229 @@
+---
+title: React Native as a monorepo
+author:
+  - Lorenzo Sciandra
+  - Tommy Nguyen
+  - Nicola Corti
+date: 2022-05-18
+---
+
+# RFC0006: React Native as a monorepo
```
cortinico, line 10: This + the file should be renamed to `0048`

kelset:
Merged 🎉

Thanks everyone who helped pushed this proposal through the finish line - it's been a massive endeavour and I want to give a special thanks to @cortinico @hoxyq and @tido64 for all their work 👏

And thanks to everyone who commented, suggested, pointed things out!

...and now we couuuuld go back to @empyrical's RFC #49 🤣 😉

Issue #512, opened 2022-09-12
Include a custom resolver in Jest preset


SimenB:
### Introduction

Metro resolves a custom field in `package.json` called `react-native`. This is not supported by Jest's default resolver - just like e.g. `module` isn't. This means different modules are resolved when running tests than when `metro` is bundling code.

### Details

The Jest preset in `react-native` could ship a [`resolver`](https://jestjs.io/docs/configuration#resolver-string) which uses e.g. `packageFilter` to prioritize `react-native` over `main` etc..

While the long term solution to this issue is probably `exports` and passing a `react-native` condition, Metro ([currently](https://github.com/facebook/metro/issues/670#issuecomment-1226005278)) doesn't support `exports` and regardless, that'll leave all existent modules in the dust.

Note that this field seems to be undocumented? Sorta mentioned [here](https://facebook.github.io/metro/docs/resolution#getpackagemainpath-string--string), but not really 😅 

Like with #509, I'd be happy to help review and/or write the code if this is accepted.

### Discussion points

Different resolution of module leads to more different code than necessary running in tests vs production code, making them more brittle and less useful. Providing a custom `resolver` in the preset could help alleviate this.

thymikee:
Metro team has plans to support `exports`. cc @motiz88 @huntie

SimenB:
Yeah, I linked to the comment where the team stated it was on the roadmap. Regardless, that'd still leave 100% of existing modules behind, so I think support for `react-native` should be added

NickGerleman:
Jest's default resolver is also missing some of the smarts for out-of-tree platforms (so the wrong platform's implementation could be included).  I remember adding a custom resolver in the RNW repo for this a while back, but IIRC @tido64 did something better as part of the rnx-kit efforts which should be publicly usable.

kelset:
Maybe you are thinking of this? https://github.com/microsoft/rnx-kit/tree/main/packages/typescript-react-native-resolver

tido64:
I think @NickGerleman is referring to https://github.com/microsoft/rnx-kit/tree/main/packages/jest-preset. It contains logic for picking out the correct platform if within an out-of-tree platform, and provides affordances for specifying platforms otherwise.

PR #534, opened 2022-10-17
RFC: Metro package exports support

huntie:
This adds a new RFC containing the intended design and rollout strategy for [package exports support](https://nodejs.org/docs/latest-v18.x/api/packages.html#package-entry-points) within Metro, which will affect how React Native apps consume npm packages and how package authors expose modules to React Native apps.

> NOTE: **This is a mostly-ready first pass** which I'm opening now in order to link related conversations to. Further commits will be stacked to address early feedback.

- [**⬇️ Near-term planned changes**](https://github.com/react-native-community/discussions-and-proposals/pull/534#issuecomment-1285311539)

---

Recommended (in the _Files changed_ tab): "Display the rich diff"
image

### DIFF
--- a/proposals/0534-metro-package-exports-support.md
+++ b/proposals/0534-metro-package-exports-support.md
@@ -1,0 +1,455 @@
+---
+title: Package exports and strict entry points support in Metro
+author:
+- Alex Hunt <alexeh@meta.com>
+date: 2022-10-17
+---
+
+# RFC0534: Package exports and strict entry points support in Metro
+
+## Summary
+
+To improve React Native’s compatibility with the npm package ecosystem, we intend to add support for the [`"exports"` field](https://nodejs.org/docs/latest-v18.x/api/packages.html#exports) in Metro. This feature was introduced in Node.js 12.7.0 and is described in the Node.js spec under [*Package entry points*](https://nodejs.org/docs/latest-v18.x/api/packages.html#package-entry-points).
+
+## Basic example
+
+The `"exports"` field is an expanded way for npm packages to define entry points within their package.json manifest. It is a modern alternative to `"main"` and the recommended method for exposing files in new packages.
+
+```json
+{
+  "name": "my-package",
+  "exports": {
+    ".": "./lib/index.js",
+    "./utils/*": "./utils/*.js",
+    "./utils/*.js": "./utils/*.js",
+    "./package.json": "./package.json"
+  }
+}
+```
+
+https://nodejs.org/docs/latest-v18.x/api/packages.html#package-entry-points
+
+With `"exports"` support enabled, Metro will need to resolve paths appropriately based on what is defined by each package. We are proposing that package maintainers will be able to opt into strict handling of `"exports"`, with lenient and backwards compatible behaviour by default.
+
+```js
+import MyPackage from 'my-package';
+// Loads ./lib/index.js
+
+import foo from 'my-package/private/foo';
+// Loads ./private/foo.js (Inaccessible under strict mode)
+```
+
+In addition to this example, `"exports"` introduces further behaviours which are discussed later in this proposal.
+
+## Motivation
+
+Developers using React Native should be able to use current JavaScript standards, conventions and packages to author their apps.
+
+React Native developers expect a seamless experience when consuming packages from npm. Supporting `"exports"` within Metro has been a specific pain point ([facebook/metro#670](https://github.com/facebook/metro/issues/670)), with this feature having been adopted by other bundlers and a growing proportion of the JavaScript ecosystem.
+
+The expected outcomes are that:
+
+1. React Native projects will work with more npm packages out-of-the-box. Packages using `"exports"` will have associated non-breaking behaviours respected under Metro (described in the next section).
+2. [React Native for Web](https://necolas.github.io/react-native-web/) projects will work with more npm packages out-of-the-box due to better-aligned resolution of the `"browser"` and `"react-native"` runtimes under `"exports"`.
+3. We will provide a migration path for React Native (and other, willing) package maintainers to opt into strict and spec-compliant handling of `"exports"` under Metro, via a breaking release.
+    - Headline benefit: Enforced encapsulation of private modules (and therefore reduced surface area for misuse/bugs).
+
+## Priorities within this proposal
+
+The Metro team intends to implement all features of `"exports"` described in this proposal. However, due to the size of the spec we are considering features in two priority levels.
+
+- **P0 - Core features**: The robust subset of features that we must support in the spec for a minimum implementation of `"exports"`.
+    - This work will be scheduled ahead of P1 features unless it makes sense not to during implementation.
+    - App developers may wish to opt in to experimental `"exports"` support upon completion of these features.
+- **P1 - Not targeted in first iteration**:
+    - Attached to features which are more challenging to implement, or that were launched in later revisions to the Node.js spec.
+    - These may still be done, otherwise will be fast-follow.
+    - This work may be up for grabs by interested external contributors.
+
+**All features discussed are P0 unless otherwise stated.**
+
+## Detailed design
+
+We intend to support all features of `"exports"` up to Node 18.10. For now, we are aware of but are not committing to support for [`"imports"`](https://nodejs.org/docs/latest-v18.x/api/packages.html#imports).
+
+We are beginning in a state where `"exports"` is in general use and we cannot guarantee that references to these packages within existing React Native projects are compliant with the `"exports"` spec. Therefore Metro needs to carefully design around this.
+
+The following features and considerations are detailed in this section:
+
+- [Package encapsulation](#package-encapsulation)
+- [Subpath exports](#subpath-exports)
+- [Subpath patterns](#subpath-patterns)
+- [Exact path specifiers](#exact-path-specifiers)
+- [Conditional exports](#conditional-exports)
+- [Conditional exports: `"import"` and `"require"`](#conditional-exports-import-and-require)
+- [Conditional exports: Community definitions and `"browser"`](#conditional-exports-community-definitions-and-browser)
+- [Conditional exports: User conditions](#conditional-exports-user-conditions)
+- [Opting into strict `"exports"` handling](#opting-into-strict-exports-handling)
+
+The following features will be implemented without any anticipated behaviour differences or special notes (links go to Node.js spec):
+
+- [Exports sugar](https://nodejs.org/docs/latest-v18.x/api/packages.html#exports-sugar)
+- [Self-referencing a package using its name](https://nodejs.org/docs/latest-v18.x/api/packages.html#self-referencing-a-package-using-its-name)
+
+### Package encapsulation
+
+This is the primary breaking change introduced by `"exports"`.
+
+> When the `"exports"` field is defined, all subpaths of the package are encapsulated and no longer available to importers.\
+– https://nodejs.org/docs/latest-v18.x/api/packages.html#main-entry-point-export
+
+```json
+"exports": {
+  // Paths outside of this list are inaccessible to importers
+  ".": "./lib/index.js",
+  "./package.json": "./package.json"
+}
+```
+
+**Proposed**:
+
+- By default, Metro will ignore this constraint. We will log a warning when an inaccessible import is accessed.
+- Under strict handling, Metro will prevent inaccessible imports and surface a resolution error to the developer.
+
+### Subpath exports
+
+This relates to the ability for packages to define subpath aliases that point to an underlying file.\
+https://nodejs.org/docs/latest-v18.x/api/packages.html#subpath-exports
+
+```
+"exports": {
+  ".": "./index.js",
+  "./submodule.js": "./src/submodule.js"
+}
+```
+
+One edge case is a path conflict, for instance if the file tree for this package looked like this:
+
+```
+├── src
+│   └── submodule.js
+└── submodule.js
+```
+
+**Proposed**:
+
+- Metro will consider all exact `"exports"` subpaths when present.
+- In the case of a path conflict, when not using strict handling, Metro will prioritise the apparent file path over the explicit subpath if the file exists.
+
+Illustration of subpath resolution for the above conflict example:
+
+| | Default handling (non-strict) | Strict handling |
+| - | - | - |
+| `'pkg/submodule.js'` | `./submodule.js` | `./src/submodule.js` |
+| `'pkg/src/submodule.js'` | `./src/submodule.js` | (Inaccessible) |
+| | *Maintains compatible behaviour* | *Always reads from `"exports"`* |
+
+### Subpath patterns
+
+Subpath exports may use patterns to match several files. These can expose all contents of a directory, or include a [*pattern trailer*](https://github.com/nodejs/node/pull/39635) (e.g. file extension).\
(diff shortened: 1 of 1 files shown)


proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,455 @@
...
+| Condition name | Description | Node.js | Webpack | Metro (proposed) |
+| - | - | - | - | - | - |
+| `"browser"` | Any web browser environment | - | Matched when `platform === 'web'` | Matched when `platform === 'web'` |
+| `"react-native"` [New] | Any React Native runtime environment (non-browser) | - | Matched when `platform !== 'web'` | Matched when `platform !== 'web'` |
+| All other | | (unimplemented) | Various supported, e.g. `"deno"`, `"worker"`, `"electron"` | (unimplemented) |
+| `"default"`| Makes no assumption about JS environment | Always matched | Always matched | Always matched |
+
+Note: Now that `"react-native"` and `"browser"` will not overlap, packages intending to support both native and web environments but which want to expose alternative `"exports"` targeting *React Native **APIs*** should continue to use the `"default"` condition and use a [subpath export](#subpath-exports) instead: e.g. `import { reactNativeExport } from 'some-pkg/react-native'`.
+
+### Conditional exports: User conditions
```
SimenB, line 308: whatever happens here needs to be accessible to the custom env (again, #509) - maybe the env should get export conditions from Metro config rather than having its own?

proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,455 @@
...
+    conditionNames: ['browser', 'production'],
+  },
+};
+```
+
+The exact name and shape of this option may change during implementation.
+
+In addition, `resolver.resolveRequest` will continue to provide an escape hatch from Metro's handling of conditional exports, should an app need to override imports from a given package.
+
+### Opting into strict `"exports"` handling
```
SimenB, line 331: if it's _not_ strict by default, #512 is probably even more needed as Jest doesn't allow you to toggle strictness (similar to Node)

proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,455 @@
...
+
+The second case above demonstrates less intuitive existing functionality within Metro, which we anticipate has not been in wide use. `"exports"` presents the opportunity to remove this.
+
+**Proposed**:
+
+- Keep current behaviour: Metro will try platform extension variants when resolving both extensionless and extensioned subpaths via `"exports"`.
+- Under strict behaviour, Metro will treat extensioned imports as direct imports and will not attempt to resolve platform extensions.
+- We will communicate to React Native package authors that extensionless exports should be preferred, since they allow the ability to introduce platform-specific modules where needed.
+
+Not committed to in this RFC: We are considering deprecating or changing support for `.native.js` (raised in [facebook/metro#874](https://github.com/facebook/metro/pull/874)). This could be coupled with `"exports"` strict behaviour.
```
tido64, line 216: I hope `.native.js` stays, otherwise I'm not aware of any other way to share a single file between multiple platforms. We have many code bases that share code between native and web. We consistently use `.native.js` for native platforms and `.js` for web.

proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,455 @@
...
+- Metro will not provide an implementation for other community conditions.
+
+Combined, this is the only breaking change we foresee in Metro that will immediately impact app developers, and which we are not deferring via strict mode opt-in. It will align behaviour for existing users of react-native-web via Webpack.
+
+| Condition name | Description | Node.js | Webpack | Metro (proposed) |
+| - | - | - | - | - | - |
+| `"browser"` | Any web browser environment | - | Matched when `platform === 'web'` | Matched when `platform === 'web'` |
+| `"react-native"` [New] | Any React Native runtime environment (non-browser) | - | Matched when `platform !== 'web'` | Matched when `platform !== 'web'` |
+| All other | | (unimplemented) | Various supported, e.g. `"deno"`, `"worker"`, `"electron"` | (unimplemented) |
+| `"default"`| Makes no assumption about JS environment | Always matched | Always matched | Always matched |
```
byCedric, line 304:
```suggestion
| Condition name | Description | Node.js | Webpack | Metro (proposed) |
| - | - | - | - | - |
| `"browser"` | Any web browser environment | - | Matched when `platform === 'web'` | Matched when `platform === 'web'` |
| `"react-native"` [New] | Any React Native runtime environment (non-browser) | - | Matched when `platform !== 'web'` | Matched when `platform !== 'web'` |
| All other | | (unimplemented) | Various supported, e.g. `"deno"`, `"worker"`, `"electron"` | (unimplemented) |
| `"default"`| Makes no assumption about JS environment | Always matched | Always matched | Always matched |
```

proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,455 @@
...
+"exports": {
+  // Paths outside of this list are inaccessible to importers
+  ".": "./lib/index.js",
+  "./package.json": "./package.json"
+}
+```
+
+**Proposed**:
+
+- By default, Metro will ignore this constraint. We will log a warning when an inaccessible import is accessed.
```
byCedric, line 111: One concern I have with these logs is that the user is not the owner of the library, and can't change the code without the library author (`unless using patch-package, but that shouldn't be a default`).

As long as the log mentions pursuing the library authors to change it, I think this is great. If that's not clear, that it's not an issue with the project itself, I think it only confuses users.

proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,455 @@
...
+  "name": "my-package",
+  "exports": {...},
+  // ...
+  "reactNative": {
+    "metroExportsMode": "strict"
+  }
+}
+```
+
+Note: `"reactNative"` (in camel case) will be a distinct field from `"react-native"` (in dash case, used as a `"main"` module alias that we are starting to deprecate with `"exports"`). We are [separately considering a `‌"newArchSupport"` option](https://github.com/microsoft/rnx-kit/issues/1863#issuecomment-1246580474) within this new top-level key, which will encapsulate options available to React Native packages.
```
huntie, line 355: @kelset @cortinico Flagging to continue the discussion from https://github.com/microsoft/rnx-kit/issues/1863.

**History**: We agreed that the suggested `codegenConfig` object in package.json could be generalised, provisionally as `reactNativeConfig`. This new key would be intended to namespace React Native config options and/or future metadata which libraries can configure.

In this PR, I'm positing the wider name of `reactNative` — and seeing what resistance there is 😅.

Additional supporting thought: Under the proposed `"metroExportsMode": "strict"`, including the `"react-native"` field will no longer be allowed by Metro — reducing the likelihood that both of these exist.

huntie:
### Near-term planned changes

*From initial internal and external feedback round. These are being strongly considered to form updated versions of this proposal.*

#### 1. Move strict handling out of current proposal

The surface area of this proposal is large and we want to be sure that package maintainers understand changes correctly and are not burdened by too many framework decisions at once (e.g. the new architecture migration is concurrently being pushed to the community).

Most of the benefits of `"exports"` relevant to React Native app developers will be adequately delivered by the proposed lenient handling by Metro, **and** `"metroExportsMode": "strict"` presents complexity and footguns both to us and to package maintainers in implementing this correctly. Therefore it makes sense not to ship `metroExportsMode` immediately.

**Proposed**: Split out all strict mode references into a future RFC (for a future effort).

#### 2. Widen definition and behaviour of `"react-native"` condition

Expand the `"react-native"` condition to represent all React Native platforms (more conceptually and community aligned), provide info on how packages should correctly use this with `"browser"` for react-native-web support.

**Proposed**: Metro will assert both conditions.

#### 3. Handling of platform-specific extensions under `"exports"`

The current plan of deciding to try extra files based on extensionless `"exports"` keys is _very_ loose behaviour and possibly spec-conflicting enough to break in other bundlers.

This could be redesigned now or deprioritised and handled as a future decision + feature — particularly as we offer alternatives for the same functionality without platform extensions.

proposals/0534-metro-package-exports-support.md
huntie: Noted. No near term plans for this change AFAIK — if so it will be stated in a future RFC.

proposals/0534-metro-package-exports-support.md
huntie: Yes we plan to make both sides of the warning clear. Provisional text for the warning is [here](https://github.com/react-native-community/discussions-and-proposals/pull/534/files#diff-40781647168e437e711b426935003e99ce4bddb4c1a5f1ebd61b0218d2b45670R399).

motiz88:
Re asserting both `"react-native"` and `"browser"` everywhere and proceeding as per spec (picking the one specified first if both are specified): I think this is the right call as the least breaking change that preserves the most future value. IMO it goes hand-in-hand with *keeping* the default `resolverMainFields: ["react-native", "browser", "main"]` setting in the CLI. The key difference is that with `exports`, *package authors* wishing to provide both `browser` and `react-native` entry points can control which one gets selected.

kelset:
hey @huntie just wanted to check in - what's the status of the work and/or of this PR? any blockers we can help with?

huntie:
@kelset No blockers as of the "Near-term planned changes" note above. I haven't started implementation yet due to internal project scheduling (have been working on another thing for the last couple of months), but I wanted to get questions answered for `"exports"` earlier and get this RFC (and Node RFC) in public.

I'm aiming to push the updated RFC doc (incorporating planned changes) to this branch next week, then seeking to merge this PR. Will ping you and others for an approval at that time.

SimenB:
FWIW, https://github.com/facebook/react-native/pull/35203 has landed, so the Jest env used in the preset uses it (or, will when published)

proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,455 @@
...
+```
+
+We believe this default is non-ideal as it inhibits packages using the `"browser"` field spec from targeting react-native-web — instead working only for React Native-specific `.web.js` extensions. We believe the majority of non-React Native community package authors (in addition to React app developers) expect `"browser"` to used on web.
+
+This default is directly misaligned with react-native-web via Webpack  ([docs](https://webpack.js.org/guides/package-exports/#target-environment)), which is used by a majority of users.
+
+**Proposed**:
+
+- **Breaking**: We will implement the `"browser"` condition name — changed from the top-level field to be selected when the React Native platform is `'web'`.
+- **Breaking**: We will introduce a `"react-native"` condition name — changed from the top-level field to represent a native React Native JS runtime environment (and not web).
```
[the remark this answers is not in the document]
huntie: Resolved in https://github.com/facebook/react-native/pull/35203. Also, see amended definition [described here](https://github.com/react-native-community/discussions-and-proposals/pull/534#issuecomment-1285311539) and submitted to npm: https://github.com/nodejs/node/pull/45367.

proposals/0534-metro-package-exports-support.md
huntie: Sounds ideal. I'd be keen if this was explicitly configured, e.g. specifying a `metroConfigPath` in `ReactNativeEnv` as a configuration point — especially if calling Metro is the solution for [aligning nonstrict exports handling in Jest (your other comment)](https://github.com/react-native-community/discussions-and-proposals/pull/534/files#r997999005).

Otherwise defining user conditions (default = `['react-native']`) in both configs as part of the project template may be adequate.

huntie:
### Changes in [1a25973](https://github.com/react-native-community/discussions-and-proposals/pull/534/commits/1a25973ca1bbe842b6c32f9cd280c911a2eb35f5)

([Planned changes above](https://github.com/react-native-community/discussions-and-proposals/pull/534#issuecomment-1285311539) were integrated in [e510faa](https://github.com/react-native-community/discussions-and-proposals/pull/534/commits/e510faaa5756dea278e32afadd6b11ee15988598).) Following this, I've pushed another revision addressing further feedback, and have ended up inverting our plans around backwards compatibility for exact path specifiers and platform-specific extensions — we would prefer to make handling of these cases breaking in order to follow the `"exports"` spec correctly (discussed with @motiz88).

Significant changes:
- Now breaking: `"exports"` subpath is preferred in a path conflict.
- Now breaking: `"exports"` subpath is used if matched, blocking platform-specific extensions resolution.

More detail on:
- Dynamic configuration of asserted condition names.
- Anticipated breaking change risk levels.
- Nested conditions support is now P0.

proposals/0534-metro-package-exports-support.md
```diff
@@ -0,0 +1,452 @@
...
+// No match in "exports" (Metro will print warning)
+// Tries files if present:
+//   pkg/src/FooComponent.[platform].js
+//   pkg/src/FooComponent.native.js
+//   pkg/src/FooComponent.js
+```
+
+The last example given is not expected to enable backwards compatibility, but may serendipitously capture pre-existing imports in apps. In a future strict mode, paths outside `"exports"` will not be considered.
+
+We will recommend that packages which rely on platform-specific extensions being available to consuming apps do not migrate to `"exports"` or change these entry points to handle platforms using `Platform.select()` or conditional exports.
```
tido64, line 261: This seems to contradict the proposed implementation of conditional exports: _"We do not intend to use conditional exports as an alternative method to target the current React Native concept of platforms (e.g. `"android"`, `"ios"`, `"web"`)."_

Can we propose that people add a public "header" that re-exports the platform-forked files, e.g.

```js
// public.js
export * from "./platform-specific";

// platform-specific.android.js

// platform-specific.ios.js
```

huntie: Partially — this refers mostly to the case of using the `"react-native"` condition to replace `.native.js` (and to a similar degree, `"browser"`). Our philosophy with this decision is to work well with packages in the wider npm ecosystem, and have slightly more opinion/restriction of patterns for React Native-specific packages.

matthew-dean:
What about [subpath imports](https://nodejs.org/api/packages.html#subpath-imports)? Is there a separate RFC for this?

proposals/0534-metro-package-exports-support.md
```diff
@@ -135,16 +135,15 @@ One edge case is a path conflict, for instance if the file tree for this package
 
 **Proposed**:
 
-- Metro will consider all exact `"exports"` subpaths when present.
-- In the case of a path conflict, when not using strict handling, Metro will prioritise the apparent file path over the explicit subpath if the file exists.
+- **Breaking**: Metro will consider all exact `"exports"` subpaths when present. In the case of a path conflict, Metro will prioritise the file specified by `"exports"`.
```
motiz88, line 138: Related: Do we specify anywhere what happens if there is a match in `exports` but it fails to resolve? (e.g. an already-published package has an `exports` field that happens to be broken, but currently works under Metro)

In non-strict mode, I think there's a case for falling back to filesystem resolution (with a warning), similar to the case where there isn't a match to begin with.

motiz88 (approved):
This is stellar work! Thanks @huntie for shepherding it through all the design discussions. LGTM, with one thing I'd love to see clarified re: path conflicts.

proposals/0534-metro-package-exports-support.md
huntie: Great point, I've amended this in 👍🏻

PR #588, opened 2023-01-25
RFC: introduce `reactNativeManifest` to package.json

kelset:
This RFC wants to introduce and formally discuss a new section for the package.json of a react-native project (app or library), called `reactNativeMetadata`.

This new section will allow developer to express certain characteristics of their code in a more formalized manner, that can easily used by tooling to act towards the code accordingly.

**Note well**: this is but the first draft of the actual shape and fields. The goal of this draft is to collect feedback and help finalise the spec for this section.

---

You can read more details in the text itself - here's the [file view](https://github.com/react-native-community/discussions-and-proposals/blob/kelset/rfc-reactNativeMeta/proposals/0012-introduce-reactNativeMetadata.md) for easier reading experience.

### DIFF
--- a/proposals/0012-introduce-reactNativeMeta.md
+++ b/proposals/0012-introduce-reactNativeMeta.md
@@ -5,7 +5,7 @@ author:
 date: 25-01-2021
 ---
 
-# RFC0011: Introducing `reactNativeMeta` to `package.json`, for RN specific metadata
+# RFC0012: Introducing `reactNativeMeta` to `package.json`, for RN specific metadata
 
 ## Summary
 
@@ -32,8 +32,8 @@ This is how this section would look like...
         "features": {
             "fabric": "true",
             "android": {
-            "turbomodules": "true",
-        },
+                "turbomodules": "true",
+            },
         },
         "requirements": {
             "react-native": ">= 0.71.0",
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -227,6 +227,7 @@ Therefore we believe `align-deps` can benefit from this information and be exten
 ### React Native CLI support
 
 Ideally the CLI (or any other build tool) should be able to consume the `reactNativeManifest` section and warn at build time if a configuration is invalid (either by having an invalid value or by having a mixture of invalid configurations).
+For example, the CLI could use [jsonschema](https://npmjs.com/package/jsonschema) at build time and validate the `package.json` against the schema we provide.
 
 It's outside the scope of this RFC to delve into details of which tool will implement which logic and we defer this to a future discussion.
 
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -134,13 +134,14 @@ The main drawback is that we'd need to convince developers to start using it, so
 
 ## Adoption strategy
 
-Once the v1.0.0 version of the shape is finalised, we can start integrating it across the board in tools such has `align-deps` and the RNDirectory, and in the react-native tools themselves. This first wave of support would have to be coordinated around a certain release that lives in RN's main branch, such as now it'd be introduced in 0.72 (which branch has not been cut).
+Once the v1.0.0 version of the shape is finalised, we can start integrating it across the board in tools such has `align-deps` and the RNDirectory, and in the react-native tools themselves. This first wave of support would have to be coordinated around a certain release that lives in RN's main branch; for example, if we were to introduce it now, it'd be in 0.72 (which branch has not been cut) via the app template.
 
 After adding this first wave of support, and related documentation, we could more broadly communicate to the community and the maintainers how adding this section would benefit them too. Especially for library maintainers we'll have another RFC coming up soon to describe how we are planning to better communicate this.
 
 ## How we teach this
 
 - add to core's documentation (which should happen anyway if we start using it for flags for new arch configuration)
+- add to the templates (both app and library)
 - have RNDirectory and align-deps mention a call to action to introduce this
 
 ## Unresolved questions
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -227,6 +227,8 @@ The React Native direcotry can easily consume the `reactNativeManifest` by:
 
 ## Future proof and extensibility
 
+We're versioning the schema using the `schema` key in the `reactNativeManifest` section. This allows us to evolve the schema in the future without breaking changes. We believe that the `schema` key should be a semver version string.
+
 We believe that the `reactNativeManifest` section should be extensible as we believe new capabilities and features can be added there.
 Specifically, we envision to evolve `reactNativeManifest` or `reactNativeManifest.capabilites` as follows:
 
@@ -246,7 +248,7 @@ For an app the section will look as follows:
   "name": "my-awesome-app",
   "version": "1.2.3",
   "reactNativeManifest": {
-    "version": "1.0",
+    "schema": "1.0.0",
     "type": "app",
     "capabilities": {
       "hermes": {
@@ -275,7 +277,7 @@ For a library the section will look as follows:
   "name": "my-awesome-library",
   "version": "1.2.3",
   "reactNativeManifest": {
-    "version": "1.0",
+    "schema": "1.0.0",
     "type": "library",
     "capabilities": {
       "hermes": {
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -105,10 +105,10 @@ reactNativeMetadata: {
             turbomodules: boolean,
             jsEngine: string,
         },
-        iOs: {
+        ios: {
         ... // same as android
         },
-        macOs: {
+        macos: {
         ... // same as android
         },
         windows: {
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -1,15 +1,15 @@
 ---
-title: Introducing `reactNativeMeta` to `package.json`, for RN specific metadata
+title: Introducing `reactNativeMetadata` to `package.json`, for RN specific metadata
 author:
   - Lorenzo Sciandra
 date: 25-01-2021
 ---
 
-# RFC0012: Introducing `reactNativeMeta` to `package.json`, for RN specific metadata
+# RFC0012: Introducing `reactNativeMetadata` to `package.json`, for RN specific metadata
 
 ## Summary
 
-This RFC wants to introduce and formally discuss a new section for the package.json of a react-native project (app or library), called `reactNativeMeta`.
+This RFC wants to introduce and formally discuss a new section for the package.json of a react-native project (app or library), called `reactNativeMetadata`.
 
 This new section will allow developer to express certain characteristics of their code in a more formalized manner, that can easily used by tooling to act towards the code accordingly.
 
@@ -26,7 +26,7 @@ This is how this section would look like...
     "name": "@rnx-kit/metro-serializer",
     "version": "1.0.11",
     ...,
-    "reactNativeMeta": {
+    "reactNativeMetadata": {
         "version": "2.3",
         "type": "library",
         "features": {
@@ -51,7 +51,7 @@ This is how this section would look like...
     "name": "contoso",
     "version": "2.0.3",
     ...,
-    "reactNativeMeta": {
+    "reactNativeMetadata": {
     "version": "1.0",
     "type": "app",
     "features": {
@@ -75,7 +75,7 @@ A similar problem was also encountered by [React Native Directory](https://react
 
 It became clear that there was a need to be able to communicate and parse certain information about a react-native library via its `package.json`, which led to [a meeting](https://github.com/microsoft/rnx-kit/discussions/2125) between Meta, Microsoft and Expo to figure out a common solution.
 
-This RFC is the result of that meeting: a new section of the `package.json`, called `reactNativeMeta`, to contain certain metadata about a react-native project - it being either a library or an app.
+This RFC is the result of that meeting: a new section of the `package.json`, called `reactNativeMetadata`, to contain certain metadata about a react-native project - it being either a library or an app.
 
 There are a few areas that this section could express in a consistent manner; here's a list of the ones we could think of (in no particular order):
 
@@ -84,14 +84,14 @@ There are a few areas that this section could express in a consistent manner; he
 - if an app has opted in into Fabric, TM, or both
 - if an app is using Hermes or another js engine
 
-Sidenote: the name `reactNativeMeta` was chosen to mirror the camel case pattern used by other fields such as `peerDependenciesMeta`.
+Sidenote: the name `reactNativeMetadata` was chosen to mirror the camel case pattern used by other fields such as `peerDependenciesMeta`.
 
 ## Detailed design
 
 This is the section that in this first draft will need to be properly finalised. For now, here's what we envisioned:
 
 ```js
-reactNativeMeta: {
+reactNativeMetadata: {
     version: string, // ex. 1.0.0 - we know that this spec might evolve in the future
     type: string; // "app" or "library"
     features: {
@@ -130,7 +130,7 @@ The main drawback is that we'd need to convince developers to start using it, so
 
 ## Alternatives
 
-`reactNativeMeta` is, by design, fully original and new: there are other files that are used to define configurations for various aspect of react-native based projects (such as `react-native.config.js`, `app.json`, Expo's `app.config.js`, `expo-module.config.json`) but this one does **not** overlap with any of them. Explicitly, this one has the purpose of being filled only with metadata to be read by package managers and other tools.
+`reactNativeMetadata` is, by design, fully original and new: there are other files that are used to define configurations for various aspect of react-native based projects (such as `react-native.config.js`, `app.json`, Expo's `app.config.js`, `expo-module.config.json`) but this one does **not** overlap with any of them. Explicitly, this one has the purpose of being filled only with metadata to be read by package managers and other tools.
 
 ## Adoption strategy
 
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -207,6 +207,17 @@ Selectively toggling Fabric/TurboModule is a more advanced feature. We believe w
 
 This section lists the tooling we believe could be built on top of `reactNativeManifest` to improve the developer experience.
 
+### Build Logic support
+
+Build tools such as Gradle/CocoaPods or others should account for the `reactNativeManifest` section and use it to configure the build pipeline accordingly to this and future RFCs. Specifically they should follow this logic for a generic capability `foo`.
+
+1. If the capability `foo` is specificed in the `reactNativeManifest.capabilities` section, use the value specified in the `reactNativeManifest.capabilities.foo` section.
+   1. If the user is **also** specifying the capability `foo` in the build tool specific configuration (e.g. inside `gradle.properties` for Android) behave as follows:
+      1. If the two values are **compatible**, use them without notifying the user.
+      1. If the two values are **incompatible**, notify the user and use the value specified in the `reactNativeManifest.capabilities.foo` section will prevail and the value specified in the build tool specific configuration will be ignored.
+   1. If the user is **not** specifying the capability `foo` in the build tool specific configuration, honor the value specified in the `reactNativeManifest.capabilities.foo` section.
+1. If the capability `foo` is **not** specificed in the `reactNativeManifest.capabilities` section, honor the value specified in the build tool specific configuration or the default value if not specified.
+
 ### `align-deps` support
 
 This RFC originated from a [conversation between Meta and Microsoft](https://github.com/microsoft/rnx-kit/issues/1863) to use `align-deps` as key feature in the New Architecture rollout support, in helping developers understanding if a library they're using is New Architecture compatible or not. Very quickly it became clear that there is currently no straightforward way to know if a library indeed has that support.
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -1,8 +1,8 @@
 ---
-title: [DRAFT] Introducing `reactNativeManifest` to `package.json` for React Native specific metadata
+title: DRAFT | Introducing `reactNativeManifest` to `package.json` for React Native specific metadata
 author:
-  - Lorenzo Sciandra
-  - Nicola Corti
+- Lorenzo Sciandra
+- Nicola Corti
 date: 25-01-2021
 ---
 
@@ -27,24 +27,24 @@ As of today, there is no simple programmatic way to know if a NPM package is a R
 Those indicators are all non-exhaustive. 
 
 Similarly, there is no easy way to understand if a library is exposing a native module or not. 
-The React Native CLI is facing this difficulty and tries to circumvent it by inspecting the native code and search for subclasses of `*ReactPackage` using complex regular expressions ([example](https://github.com/react-native-community/cli/blob/605c542d18efcb02f217d3c85726fa73a79054c2/packages/cli-platform-android/src/config/findPackageClassName.ts#L35)) due to the lack of a declarative way to express this information.
+The React Native CLI is facing this difficulty and tries to circumvent it by inspecting the native code and searching for subclasses of `*ReactPackage` using complex regular expressions ([example](https://github.com/react-native-community/cli/blob/605c542d18efcb02f217d3c85726fa73a79054c2/packages/cli-platform-android/src/config/findPackageClassName.ts#L35)) due to the lack of a declarative way to express this information.
 
-Another practical examples of this need are the [React Native Directory](https://reactnative.directory/) attempting to [programmatically infer](https://github.com/react-native-community/directory/pull/870) New Architecture support from the presence of the `codegenConfig` fields in the `package.json` of a library. The `codegenConfig` filed is not intended to carry semantic information about New Architecture support, and is definitely a non-exhaustive solution for this problem.
+Another practical example of this need is the [React Native Directory](https://reactnative.directory/) attempting to [programmatically infer](https://github.com/react-native-community/directory/pull/870) the New Architecture support from the presence of the `codegenConfig` fields in the `package.json` of a library. The `codegenConfig` field is not intended to carry semantic information about New Architecture support, and is definitely a non-exhaustive solution for this problem.
 
-While we can continue to build tooling that tries to infer metadata and capabilities of a library, providing a declarative way to declared such capabilites provides a series of benefit:
+While we can continue to build tooling that tries to infer metadata and capabilities of a library, providing a declarative way to describe such capabilities yields a series of benefit:
 * **predictability**: developers can easily understand what a library is capable of doing and if such capability matches their app.
 * **tooling**: tooling can easily consume this information and provide a better experience to developers.
 * **single source of thruth**: a single manifest will act as a source of truth for all those capabilities and flag definition, that can easily be consumed by different platforms to provide a unified way to enable/disable capabilities.
-* **verifiability**: the manifest can be easily verified & linted by tooling to ensure that app & library are not using an incompatible set of compatibilites. Similarly, the JSON schema we publish can be used to validate the `reactNativeManifest` section.
+* **verifiability**: the manifest can be easily verified and linted by tooling to ensure that apps and libraries are not using an incompatible set of dependencies. Similarly, the JSON schema we publish can be used to validate the `reactNativeManifest` section.
 
 ## Use Cases
 
-This section lists the use case we envision for the `reactNativeManifest` section.
+This section lists the use cases we envision for the `reactNativeManifest` section.
 
 ### Declaring the type of a project
 
 For the sake of this RFC, we consider a React Native project as either an app or a library.
-While an app project can contain native modules and components, it's **not** possible for a app to "consume" another app as a library as this would break the native build system invariants (i.e. a single `AndroidManifest.xml` and `Info.plist` file per app, etc.). If an app projects wishes to distribute some of its native modules or components, they need to be extracted into a library project.
+While an app project can contain native modules and components, it's **not** possible for an app to "consume" another app as a library. This would break the native build system invariants (i.e. a single `AndroidManifest.xml` and `Info.plist` file per app, etc.). If an app projects wishes to distribute some of its native modules or components, they need to be extracted into a library project.
 
 We're proposing to declare if a project is either a library or an app by adding the `type` field (as `"app"` or `"library"`) to the `reactNativeManifest` section of the `package.json` file.
 
@@ -59,19 +59,19 @@ We're proposing to declare if a project is either a library or an app by adding 
 ### New Architecture support
 
 One of the primary driver of this proposal is the [New React Native Architecture support](https://reactnative.dev/docs/the-new-architecture/landing-page). Specifically we believe that the ecosystem is currently affected by:
-* Lack of a declarative way to define if a library is compatible with New Architecture
-  * As described in the [rationale](#rationale), tooling needs to infer this information from the presence of the `codegenConfig` field in the `package.json` of a library. Libraries might decide to don't use codegen and still be compatible with New Architecture, so this method is not exhaustive.
-  * The alternative at this stage is to inspect the code and check the API callsites if they are New Architecture or Old Architecture compatible.
-* Lack of a declarative way to enable New Architecture on both platforms
-  * Currently New Architecture is enabled by a Gradle Property `newArchEnabled` on Android in the `gradle.properties` file and by invokign `RCT_NEW_ARCH_ENABLED pod install` on iOS.
-  * As of today, there is no way to enable New Architecture for an app project for both platforms
-  * Moreover, the file vs environment variable leads to a scenario where you can't know if a app supports New Architecture or not by inspecting the code (as New Architecture support is known at  `pod install` time and is not codified in the codebase)
+* Lack of a declarative way to define if a library is compatible with the New Architecture
+  * As described in the [rationale](#rationale), tooling needs to infer this information from the presence of the `codegenConfig` field in the `package.json` of a library. Libraries might decide not to use Codegen and still be compatible with the New Architecture, so this method is not exhaustive.
+  * The alternative at this stage is to inspect the code and check whether the API callsites are New Architecture or Old Architecture compatible.
+* Lack of a declarative way to enable the New Architecture on both platforms
+  * Currently the New Architecture is enabled by a Gradle Property `newArchEnabled` on Android in the `gradle.properties` file and by invoking `RCT_NEW_ARCH_ENABLED=1 bundle exec pod install` on iOS.
+  * As of today, there is no way to enable the New Architecture for an app project for both platforms
+  * Moreover, the difference in how the New Architecture is enabled leads to a scenario where you can't statically know if an app supports New Architecture or not (as the New Architecture support is known at  `pod install` time and is not codified in the codebase)
 
 Therefore we propose to add the `newArch` section to the `reactNativeManifest.capabilities` of **both apps and libraries** with the following semantic:
-* For Apps: `newArch.enabled==true` implies that the app wants to use the New Architecture.
-* For Libraries: `newArch.enabled==true` implies that the library is compatible with the New Architecture.
+* For Apps: `newArch.enabled==true` means that the app wants to use the New Architecture.
+* For Libraries: `newArch.enabled==true` means that the library is compatible with the New Architecture.
 
-Tools can be built on top of this information to check that a app with `newArch.enabled==true` is not accepting libraries with `newArch.enabled==false` and warning against library that don't have the key specified.
+Tools can be built on top of this information to check that an app with `newArch.enabled==true` is not accepting libraries with `newArch.enabled==false` and to warn against library that don't have the key specified.
 
 The setup would look as follows for both apps and libraries:
 
@@ -87,7 +87,7 @@ The setup would look as follows for both apps and libraries:
 }
 ```
 
-This section will allows also for split configuration between platforms:
+This section will allow also for split configuration between platforms:
 
 ```json
 {
@@ -112,7 +112,7 @@ This section will allows also for split configuration between platforms:
 
 #### `codegenConfig` support
 
-Similarly to New Architecture support metadata, the `codegenConfig` is a key metadata of the New Architecture build pipeline.
+Similarly to the New Architecture support metadata, the `codegenConfig` is a key metadata of the New Architecture build pipeline.
 Currently the `codegenConfig` is a **top-level** key in the `package.json` of a project. 
 
 We propose to move this key under the `reactNativeManifest.capabilites` section as follows:
@@ -135,7 +135,7 @@ We propose to move this key under the `reactNativeManifest.capabilites` section 
 
 ### Hermes Support
 
-Similarly to New Architecture support, the current way to enable/disable the Hermes engine is toggled by using a Gradle Property `hermesEnabled` on Android and changing the `:hermes_enabled` property in the Podfile.
+Similarly to the New Architecture support, the current way to enable/disable the Hermes engine is toggled by using a Gradle Property `hermesEnabled` on Android and changing the `:hermes_enabled` property in the Podfile.
 
 We propose to add the `hermes` section under the `reactNativeManifest.capabilities` section as follows:
 
@@ -200,8 +200,8 @@ For the time being, we're not planning to add a `version` section to the `reactN
 
 ### TurboModule/Fabric toggles
 
-At the time of writing, we prefer to don't offer a dedicated section to toggle Fabric/TurboModule capability inside `reactNativeManifest`. The rationale is that we believe that the `newArch` section is sufficient to expose the New Architecture to users.
-Selectively toggling Fabric/TurboModule is a more advanced feature. We believe we'll still be offering a more advanced way to feature toggle such infrastructure pieces, but at the current state, our preference is to not expose such capability in the top level `reactNativeManifest` section.
+At the time of writing, we prefer not to offer a dedicated section to toggle Fabric/TurboModule capability inside `reactNativeManifest`. The rationale is that we believe that the `newArch` section is sufficient to expose the New Architecture to users.
+Selectively toggling Fabric/TurboModule is a more advanced feature. We believe we'll still be offering a more advanced way to enable or disable those pieces of the New Architecture infrastructure, but at the current state, our preference is not to expose such capability in the top level `reactNativeManifest` section.
 
 ## Proposed Tooling
 
@@ -220,9 +220,9 @@ Build tools such as Gradle/CocoaPods or others should account for the `reactNati
 
 ### `align-deps` support
 
-This RFC originated from a [conversation between Meta and Microsoft](https://github.com/microsoft/rnx-kit/issues/1863) to use `align-deps` as key feature in the New Architecture rollout support, in helping developers understanding if a library they're using is New Architecture compatible or not. Very quickly it became clear that there is currently no straightforward way to know if a library indeed has that support.
+This RFC originated from a [conversation between Meta and Microsoft](https://github.com/microsoft/rnx-kit/issues/1863) to use `align-deps` as key feature in the New Architecture rollout support, in helping developers understand if a library they're using is compatible with the New Architecture or not. Very quickly it became clear that there is currently no straightforward way to know if a library indeed has that support.
 
-Therefore we believe `align-deps` can benefit from this information and be extended to offer New Architecture support information.
+Therefore we believe `align-deps` can benefit from this information and can be extended to offer New Architecture support information.
 
 ### React Native CLI support
 
@@ -247,7 +247,7 @@ Specifically, we envision to evolve `reactNativeManifest` or `reactNativeManifes
 1. Each addition of a new capability [requires a new RFC](https://github.com/react-native-community/discussions-and-proposals#proposals) to be approved.
 2. Each RFC should define a new section in the `reactNativeManifest` with what's the intended use case.
 3. Each RFC should define what are the sensible defaults and should refrain from breaking changes.
-4. Each RFC should define the deprecation strategy for previous flags/features related to the same capability.
+4. Each RFC should define the deprecation strategy for previous flags/features related to the same capability, if needed.
 
 ## Complete Example
 
@@ -324,7 +324,7 @@ The JSON Schema will be published as soon as we find an agreement on the key nam
 If we decide to adopt this proposal, we'll have to broadcast this change to the ecosystem.
 Ideally, we'll start by adding the `reactNativeManifest` section to the app template and to `create-react-native-library`.
 
-We foresee a phase in which only a part of the userbase has this section configured and tool developers need to account for the absence of this section.
+We foresee a phase in which only a part of the userbase has this section configured, and tool and developers need to account for the absence of this section.
 
 Once the v1.0.0 version of the shape is finalised, we can start integrating it across the board in various tools mentioned in [Proposed Tooling](#proposed-tooling) section. This first wave of support would have to be aligned around a given React Native release, say in 0.73 (which branch has not been cut).
 
--- a/proposals/0012-introduce-reactNativeMetadata.md
+++ b/proposals/0012-introduce-reactNativeMetadata.md
@@ -1,149 +1,329 @@
 ---
-title: Introducing `reactNativeMetadata` to `package.json`, for RN specific metadata
+title: [DRAFT] Introducing `reactNativeManifest` to `package.json` for React Native specific metadata
 author:
   - Lorenzo Sciandra
+  - Nicola Corti
 date: 25-01-2021
 ---
 
-# RFC0012: Introducing `reactNativeMetadata` to `package.json`, for RN specific metadata
+# RFC0588: [DRAFT] Introducing `reactNativeManifest` to `package.json` for React Native specific metadata
 
 ## Summary
 
-This RFC wants to introduce and formally discuss a new section for the package.json of a react-native project (app or library), called `reactNativeMetadata`.
+> **Warning** ⚠️ This is the initial draft of this proposal. The goal of this draft is to collect feedback and help finalize the spec for this section.
 
-This new section will allow developer to express certain characteristics of their code in a more formalized manner, that can easily used by tooling to act towards the code accordingly.
+This RFC introduces and formally discuss the concept of _React Native Manifest_, a set of metadata about a React Native project (either app or library). This metadata can be consumed by tooling to enhance the developer experience and support the React Native ecosystem.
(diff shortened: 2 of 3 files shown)


proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+author:
+  - Lorenzo Sciandra
+date: 25-01-2021
+---
+
+# RFC0012: Introducing `reactNativeMeta` to `package.json`, for RN specific metadata
+
+## Summary
+
+This RFC wants to introduce and formally discuss a new section for the package.json of a react-native project (app or library), called `reactNativeMeta`.
```
cortinico, line 12: I think the naming will create confusion with our internal stuff so `reactNativeMetadata` would be better.

As React Native is developed by Meta, the `reactNativeMeta` field might give the false impression that this is something related to our internal infrastructure which I believe we should avoid.

cortinico (commented):
Thanks for the writeup @kelset 
I'll do a further review in the next days. I'll just left one thoughts I had on top of my mind

proposals/0012-introduce-reactNativeMeta.md
tido64: Just to clarify, this name was chosen because there's prior art: [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#peerdependenciesmeta).

> might give the false impression that this is something related to our internal infrastructure

I honestly don't think this will be an issue if the documentation says to use it.

cortinico: > Just to clarify, this name was chosen because there's prior art: [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#peerdependenciesmeta).

Thanks for sharing this (I was actually not aware of `peerDependenciesMeta`). I could see `reactNativeMeta` as a valid naming if we were to have a `reactNative` field as well, and this new field provides supporting metadata to the previous configuration. Given that we don't have a `reactNative` config, I'd prefer `reactNativeMetadata` here. I think we have little value in keeping `reactNativeMeta` aligned to `peerDependenciesMeta`.

tido64: > > Just to clarify, this name was chosen because there's prior art: [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#peerdependenciesmeta).
> 
> Thanks for sharing this (I was actually not aware of `peerDependenciesMeta`). I could see `reactNativeMeta` as a valid naming if we were to have a `reactNative` field as well, and this new field provides supporting metadata to the previous configuration. Given that we don't have a `reactNative` config, I'd prefer `reactNativeMetadata` here. I think we have little value in keeping `reactNativeMeta` aligned to `peerDependenciesMeta`.

I'm not particularly attached to the name, and I don't really care what we end up choosing. But I don't quite follow this logic. "Meta" is just a shorter name for "Metadata". If `reactNativeMeta` shouldn't exist because `reactNative` doesn't, then neither should `reactNativeMetadata`.

NickGerleman: IMO it should just be `reactNative`. `peerDependenciesMeta` kinda makes sense in the world where `peerDependencies` already exists and it well-formed, a new way to extend them was needed. But there are plenty of `package.json` config keys already, and tools like `cosmicconfig` make that pattern common. E.g. `jest: {}` or `beachball: {}` in the package.

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+
+...for a library:
+
+```json
+{
+    "name": "@rnx-kit/metro-serializer",
+    "version": "1.0.11",
+    ...,
+    "reactNativeMeta": {
+        "version": "2.3",
```
NickGerleman, line 30: Most package.json extensions for configuration I have seen are not versioned. When I initially read this, I couldn't tell it was meant to be a schema version until I looked lower.

Easier said than done... but would it be sufficient to try to avoid breaking existing fields, and just being additive with new ones? There are definitely projects with more complex configs which have managed that over multiple years.

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+        "type": "library",
+        "features": {
+            "fabric": "true",
+            "android": {
+                "turbomodules": "true",
+            },
+        },
+        "requirements": {
+            "react-native": ">= 0.71.0",
+            "expo": ">= 46.0.0",
```
NickGerleman, line 40: What extra information does this give over peerDependencies? Related to that, there may be cases where platforms are optional, vs not (e.g. a module may work on vs require a certain platform), which we may want to express.

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+}
+```
+
+## Drawbacks
+
+The main drawback is that we'd need to convince developers to start using it, so we'd have to expect a phase in which only a part of the userbase has this section configured and tool developers need to account for the section missing entirely.
+
+## Alternatives
+
+`reactNativeMeta` is, by design, fully original and new: there are other files that are used to define configurations for various aspect of react-native based projects (such as `react-native.config.js`, `app.json`, Expo's `app.config.js`, `expo-module.config.json`) but this one does **not** overlap with any of them. Explicitly, this one has the purpose of being filled only with metadata to be read by package managers and other tools.
```
NickGerleman, line 133: As this grows, or if we want to allow it to be dynamic, or extendable, it seems beneficial to allow mirroring into a `react-native.config.js` outside the direct package.json (maybe not that since already taken).  E.g. see recent template change moving Jest config out of package.json (or... cosmic-config bits).

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+    type: string; // "app" or "library"
+    features: {
+        // can be global...
+        newArchEnabled: boolean,
+        jsEngine: string, // ex. "hermes", "jsc", "V8"...
+        // ...or platform specific
+        android: {
+            newArchEnabled: boolean,
+            fabric: boolean,
+            turbomodules: boolean,
```
NickGerleman, line 105: nit for this specific example, having both `newArchEnabled` and others separately can create unclear states. E.g. what does it mean to have newArchEnabled false, but fabric enabled? Would be clearer to just have TM + Fabric imo.

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+    requirements: {
+        "react-native": string, // semver compliant range of versions of RN supported
+        expo: string, // semver compliant range of versions of Expo supported (if any)
+    }
+}
+```
+
+## Drawbacks
+
+The main drawback is that we'd need to convince developers to start using it, so we'd have to expect a phase in which only a part of the userbase has this section configured and tool developers need to account for the section missing entirely.
```
NickGerleman, line 129: App adoption is easier, since we can just stick it in the template if we want to align on it. Library adoption is trickier, but we would provide initial exposure at least.

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+        newArchEnabled: boolean,
+        jsEngine: string, // ex. "hermes", "jsc", "V8"...
+        // ...or platform specific
+        android: {
+            newArchEnabled: boolean,
+            fabric: boolean,
+            turbomodules: boolean,
+            jsEngine: string,
+        },
+        iOs: {
```
NickGerleman, line 108: nit: capitalization

For existing config, we use the same platform extensions as you would during bundling. All lowercase.

proposals/0012-introduce-reactNativeMeta.md
NickGerleman: Edit: Ah, I can see this providing more specificity when allowed per-platform.

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+
+The main drawback is that we'd need to convince developers to start using it, so we'd have to expect a phase in which only a part of the userbase has this section configured and tool developers need to account for the section missing entirely.
+
+## Alternatives
+
+`reactNativeMeta` is, by design, fully original and new: there are other files that are used to define configurations for various aspect of react-native based projects (such as `react-native.config.js`, `app.json`, Expo's `app.config.js`, `expo-module.config.json`) but this one does **not** overlap with any of them. Explicitly, this one has the purpose of being filled only with metadata to be read by package managers and other tools.
+
+## Adoption strategy
+
+Once the v1.0.0 version of the shape is finalised, we can start integrating it across the board in tools such has `align-deps` and the RNDirectory, and in the react-native tools themselves. This first wave of support would have to be coordinated around a certain release that lives in RN's main branch, such as now it'd be introduced in 0.72 (which branch has not been cut).
```
NickGerleman, line 137: For the app config, would the idea be to also teach native build logic how to read the flags?

proposals/0012-introduce-reactNativeMeta.md
```diff
@@ -0,0 +1,148 @@
...
+This is the section that in this first draft will need to be properly finalised. For now, here's what we envisioned:
+
+```js
+reactNativeMeta: {
+    version: string, // ex. 1.0.0 - we know that this spec might evolve in the future
+    type: string; // "app" or "library"
+    features: {
+        // can be global...
+        newArchEnabled: boolean,
+        jsEngine: string, // ex. "hermes", "jsc", "V8"...
```
NickGerleman, line 100: FWIW, the current cases using V8 are more dynamic than per-project. So they wouldn't end up using this sort of flag.

proposals/0012-introduce-reactNativeMeta.md
kelset: @NickGerleman we do not want to use `reactNative`, it will lead to confusion as there's already a `react-native` usage in `package.json` because of some web/bundler related stuff, see more https://github.com/nodejs/node/pull/45367

proposals/0012-introduce-reactNativeMeta.md
kelset: it's also complicated by:
- npm / yarn differences (npm 7 auto install)
- Tree flattening algorithms for npm / yarn / workspaces

@brentvatne can provide more details

proposals/0012-introduce-reactNativeMeta.md
kelset: `react-native.config.js` already exists and is used for CLI stuff (see https://github.com/react-native-community/cli/blob/main/docs/configuration.md#configuration) - it is an explicit goal of this metadata field to be entirely separate.

proposals/0012-introduce-reactNativeMeta.md
kelset: yes this section can be clarified and expanded to also mention how this would show up in the template, I'll do another pass tomorrow to add more info also based on other comments.

proposals/0012-introduce-reactNativeMeta.md
kelset: given how I'd expect a good portion of the fields to be related to the new architecture, and how things might vary (think of the codegen) between each version, I think it's going to be very likely that things will change in a "breaking change" way (reason why new arch is still experimental) - and also I can see a future in which a new major version will have removed all references to the new arch as new arch but simply as the "arch we use"

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
+---
+title: Introducing `reactNativeMetadata` to `package.json`, for RN specific metadata
+author:
+  - Lorenzo Sciandra
+date: 25-01-2021
+---
+
+# RFC0012: Introducing `reactNativeMetadata` to `package.json`, for RN specific metadata
```
cortinico, line 8:
```suggestion
# RFC0588: Introducing `reactNativeMetadata` to `package.json`, for RN specific metadata
```

I believe this was discussed in the past to use PR numbers as rfc number

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+    "version": "1.0",
+    "type": "app",
+    "features": {
+        "newArchEnabled": "true",
+        "android": {
+            "codegen": "true",
+            "turbomodules": "true",
+        },
+        // Future example
+        // "metroExportsMode": "strict"
```
cortinico, line 64: nit: Can we make this a valid JSON as Github renders this in a terrible way:
Screenshot 2023-01-27 at 18 48 49

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+    },
+    ...
+    }
+```
+
+## Motivation
+
+The conversation that originated this idea was [an ask from the Meta team](https://github.com/microsoft/rnx-kit/issues/1863) to have `align-deps` help developers in the community migrating to the new architecture by providing insights into which libraries would have Fabric/TM support. Very quickly it became clear that there is currently no straightforward way to know if a library indeed has that support.
+
+A similar problem was also encountered by [React Native Directory](https://reactnative.directory/), that tried to [programmatically infer](https://github.com/react-native-community/directory/pull/870) support via the presence of certain fields in the package.json of a library - this because one can get access to the manifest file via npm without having to download the entire node_module.
```
cortinico, line 74:
```suggestion
A similar problem was also encountered by [React Native Directory](https://reactnative.directory/), that tried to [programmatically infer](https://github.com/react-native-community/directory/pull/870) support via the presence of certain fields in the package.json of a library - this because one can get access to the manifest file via npm without having to download the entire NPM package.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+}
+```
+
+## Drawbacks
+
+The main drawback is that we'd need to convince developers to start using it, so we'd have to expect a phase in which only a part of the userbase has this section configured and tool developers need to account for the section missing entirely.
+
+## Alternatives
+
+`reactNativeMetadata` is, by design, fully original and new: there are other files that are used to define configurations for various aspect of react-native based projects (such as `react-native.config.js`, `app.json`, Expo's `app.config.js`, `expo-module.config.json`) but this one does **not** overlap with any of them. Explicitly, this one has the purpose of being filled only with metadata to be read by package managers and other tools.
```
cortinico, line 133: > to be read by package managers and other tools.

I think we should add here what should be the semantic if say `jsEngine: "hermes"` is defined in `package.json` but the `ios/Podfile` or `android/app/gradle.properties` file are disagreeing.

I believe `package.json` values should take higher priority above everything.

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+    requirements: {
+        "react-native": string, // semver compliant range of versions of RN supported
+        expo: string, // semver compliant range of versions of Expo supported (if any)
+    }
+}
+```
+
+## Drawbacks
+
+The main drawback is that we'd need to convince developers to start using it, so we'd have to expect a phase in which only a part of the userbase has this section configured and tool developers need to account for the section missing entirely.
```
cortinico, line 129: nit: Is it worth investigating a way to use the directory to "distribute" pull requests to known libraries to introduce this block (or have a community effort around it to do it manually)?

proposals/0012-introduce-reactNativeMeta.md
cortinico: +1 to this. I'm leaning towards having only `newArchEnabled` at the top level

proposals/0012-introduce-reactNativeMeta.md
NickGerleman: Could you give an example of the sort of breaking change you'd want to make?

proposals/0012-introduce-reactNativeMeta.md
NickGerleman: Call it `react-native-meta.config.js` as an example. But it seems like the app fields are being used for configuration. If that configuration becomes non-trivial, it may scale badly inlined into each package.json vs being shareable. The mapping of `foo` in package.json to `foo.config.js` is also somewhat expected by convention imo.

NickGerleman: > filled only with metadata to be read by package managers and other tools

What does this mean? It seems like it allows it to be used for almost anything.

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+
+```json
+{
+    "name": "@rnx-kit/metro-serializer",
+    "version": "1.0.11",
+    ...,
+    "reactNativeMetadata": {
+        "version": "2.3",
+        "type": "library",
+        "features": {
```
NickGerleman, line 32:
```suggestion
        "environments": {
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+            "fabric": "true",
+            "android": {
+                "turbomodules": "true",
+            },
+        },
+        "requirements": {
```
NickGerleman, line 38:
```suggestion
        "allowedDependencies": {
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+
+```json
+{
+    "name": "contoso",
+    "version": "2.0.3",
+    ...,
+    "reactNativeMetadata": {
+    "version": "1.0",
+    "type": "app",
+    "features": {
```
NickGerleman, line 57: I'm not sure that `features` here makes as much sense as it does in the library scenario.

NickGerleman, line 57:
```suggestion
    "build": {
```

proposals/0012-introduce-reactNativeMetadata.md
NickGerleman, line 64: If this is for an app, should the metro configuration be happening in the `metro.config.js`?

motiz88:
> package.json of a react-native project (app or library), called reactNativeMetadata

High-level comment: I don't think solving for apps and libraries with the same mechanism is the right call. `package.json` should ideally contain *only* configuration/metadata that applies to the code within that package, makes sense in the published version of the package, and is respected by all relevant tools when the package is transitively included in an app project. App-level config should live elsewhere (and probably already does).

kelset:
hey folks - quick update here: there have been some conversations happening behind the scenes but current goal is to get to a new level of alignment on the topic and update the PR here by end of next week

motiz88:
I've spent some time reading this RFC and talking to @cortinico, and I've identified my main issues with it. I'll start with the more focused points, with broader concerns towards the end.

## Autocomplete / IDE support

If we add a field to `package.json` as opposed to a separate file, what are our options for providing a schema that developers can plug into their IDE to get autocompletion etc? Do IDEs like VS Code offer ways to compose extended schemas onto the base `package.json`? (Any DSL we invent should have a good authoring experience as a baseline expectation, so we should design for this even if it's not the top priority.)

 ## Naming

This is a small & bikesheddy point, but I'm gravitating towards *`reactNativeManifest`* as the name, inspired by Android manifest files. (And hey, at least I'm not proposing XML. 😅) In my view the term "metadata" can range from human-readable and imprecise to machine-readable and well-defined; the name should convey that this is more the latter, and I feel like "manifest" does it a bit better than "metadata".

## Scope creep (current and future)

A "bag of properties" config paradigm can easily end up confusing people by conflating properties that apply at different times, to different scopes, and have different levels of strictness / user-friendliness. In short, we need to ensure this mechanism stays cohesive and easy to learn. We should apply a lot of scrutiny to potential use cases here and lay down stronger guidelines on what's a good fit for `reactNativeMetadata` vs some other mechanism.

The current proposal introduces `reactNativeMetadata` with the following description:

> This new section will allow developer to express certain characteristics of their code in a more formalized manner, that can easily used by tooling to act towards the code accordingly.

I think that's too open-ended. I'd propose wording to the effect of:

> `reactNativeMetadata` is an optional object in `package.json` describing specific aspects of the package that are unique to integrating with React Native, which cannot be inferred from standard fields in `package.json`. It's mainly intended to enable package discovery, autolinking and build-time validation in React Native projects. The metadata is intended to be useful both for *published* packages (e.g. on npm) and private/workspace packages. The scope of the metadata extends only to files included in the package, excluding any files under `./node_modules`.

Concretely:

* `metroExportsMode` *does not* fit here: it's not specific to integrating with React Native (as Metro can also be used outside of React Native). I propose that if we want to create this flag in the future, we leave it outside of `reactNativeMetadata`
* `requirements` seems redundant given that `peerDependencies` exists ([discussion](https://github.com/react-native-community/discussions-and-proposals/pull/588#discussion_r1087186285)). Even if there are theoretical benefits, they will be hard to teach and maintain. I propose that we remove this and separately explore using `peerDependencies` (in combination with `peerDependenciesMeta`?) to solve the same problem.
* I would suggest that we specify here that the semantics/shape of `reactNativeMetadata` can only change further via a follow-up RFC (and tool-specific extensions are strictly disallowed).

## What is the big picture design?

**tl;dr for the purposes of this RFC:** we should *not* block on solving this much larger challenge, BUT we should ruthlessly limit the scope to what's *immediately useful* so that we can focus future efforts on a more comprehensive solution.

To put it bluntly, it feels like we're dancing around the lack of an end-to-end, user-extensible, *designed*, build system for native code in React Native projects. We do have a patchwork of systems in place, that many people are using successfully, but the ad hoc nature of these solutions leaves us with a poor caching/incremental builds story, scattered documentation, duplication of tools/effort, wildly differing levels of polish, *and* - yes - an unclear source of truth for build configuration in a project.

AFAICT, `reactNativeMetadata` aims to partially address the "source of truth" problem, which is great - but I worry that we may be adopting constraints that wouldn't really work for a more fundamental solution. For example, the use of JSON to express what is essentially *build logic* (which even includes conditionals, i.e. platforms) seems short-sighted from this perspective. The more complexity we add down the line, the more we will start straining against the lack of Turing-complete facilities and other features of mature build systems. (Or worse: we will start adding them in... badly. See [Greenspun's tenth rule](https://en.wikipedia.org/wiki/Greenspun%27s_tenth_rule).)

**Concretely**, we should put resources towards designing the overall build system for React Native, and think of our current efforts as the first few steps towards that broader vision.

That cohesive system *may or may not be based on Buck*; it's much too early to take a firm stance either way. (And frankly this idea might not even materialise soon - for now I'm mainly hoping to start a discussion around it.) But the concept of specifying a single cross-platform graph of build actions, encompassing all the configuration, codegen, compilation, linking, and bundling in a React Native project, is something that Buck makes relatively natural - so I think it's one of the options we should evaluate more closely.

To be clear, there are tough design problems here, e.g. how the unified build system should interoperate with Xcode / Android Studio / and their underlying build systems & package managers - and I'm not claiming to have complete answers at this stage. But there are encouraging precedents at Meta for tools that do glue Buck with Xcode and Android Studio while providing a good and consistent developer experience.

liamjones:
> If we add a field to package.json as opposed to a separate file, what are our options for providing a schema that developers can plug into their IDE to get autocompletion etc? Do IDEs like VS Code offer ways to compose extended schemas onto the base package.json schema? (Any DSL we invent should have a good authoring experience as a baseline expectation, so we should design for this even if it's not the top priority.)

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,149 @@
...
+
+This is the section that in this first draft will need to be properly finalised. For now, here's what we envisioned:
+
+```js
+reactNativeMetadata: {
+    version: string, // ex. 1.0.0 - we know that this spec might evolve in the future
+    type: string; // "app" or "library"
+    features: {
+        // can be global...
+        newArchEnabled: boolean,
```
johnf, line 99: Would this be better future-proofed by doing something similar to ```jsEngine```

```suggestion
        architecture: "new", // ex. "new", "old", "futureMagicalArch"
```

cortinico:
Hey all,
I know this RFC was silent for a bit. We reworked it quite substantially to reduce the scope a bit and make it more focused on supporting capabilities. I would be supporting @kelset in driving this forward and I'd like to get another round of review

proposals/0012-introduce-reactNativeMeta.md
brentvatne: peer dependencies brings in too many we don't want, like those listed by @kelset above. here's one example of how peer can break things pretty easily, and why I recommend not using peer dependencies for react-native in general:

let's say you are building a library that is built on some react-native and expo primitives that are relatively stable, like view and text from react-native, and so your library expresses a peer dependency as follows:

```
{
  "name": "react-native-peer-dependencies",
  "version": "1.0.0",
  "main": "index.js",
  "peerDependencies": {
    "react-native": ">=0.70.0",
    "expo": ">=47.0.0"
  }
}
```

now @kelset and the release crew release `react-native@0.71.0-rc.0` and you go to install it. oh no! that does not work, because peer dependencies do not work with any beta or rc versions. [npm has no intention to change this behavior](https://github.com/npm/npm/issues/8854). 

```
npm install react-native-peer-dependencies
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: example-app@1.0.0
npm ERR! Found: expo@47.0.0-beta.1
npm ERR! node_modules/expo
npm ERR!   expo@"47.0.0-beta.1" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer expo@">=47.0.0" from react-native-peer-dependencies@1.0.0
npm ERR! node_modules/react-native-peer-dependencies
npm ERR!   react-native-peer-dependencies@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR! See /Users/brentvatne/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/brentvatne/.npm/_logs/2023-04-03T23_36_02_890Z-debug-0.log
```

brentvatne:
> * `requirements` seems redundant given that `peerDependencies` exists ([discussion](https://github.com/react-native-community/discussions-and-proposals/pull/588#discussion_r1087186285)). Even if there are theoretical benefits, they will be hard to teach and maintain. I propose that we remove this and separately explore using `peerDependencies` (in combination with `peerDependenciesMeta`?) to solve the same problem.

I want to push back against this point - the `peerDependencies` field brings a lot of baggage with it, and in the React Native ecosystem it often leads to a bad user experience in the form of unactionable/unhelpful error messages and obscure bugs related to package hoisting or duplicate versions of packages installed, where only a single version actually ever makes sense in the project (eg: for native modules, where only one version can be linked at any given time).

That said, I do appreciate your point about tightening this specification up and I think we could simplify the `requirements` field by allowing only specific keys instead of arbitrary package names. I think the most important ones to support are `react-native` and `expo`. This would allow us to recommend that library authors express peer dependencies on `react-native` and `expo` using the `*` matcher, which indicates they are required but otherwise mostly skips the default package manager peer dependency validation (this validation would break with alpha/beta/rc releases) and delegates those validations to tools that we build instead. We can use this information for a few purposes:

- Show version compatibility on https://reactnative.directory
- Validate project dependencies against that data
- Upgrade package versions automatically when upgrading React Native versions based on that data
- Installing correct versions of packages for the given React Native / Expo versions

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+This new section will allow developers to follow a declarative approach to express capabilities of apps and libraries in a more formalized manner.
+
+## Rationale
+
+The need to specify **capabilities and metadata** for a react-native library is inherent to the React Native ecosystem.
+
+As of today, there is no simple programmatic way to know if a NPM package is a React Native library. Good indicators could be:
+* Presence of a `react-native:*` dependency in the `peerDependencies` section of the `package.json` ([example](https://github.com/RonRadtke/react-native-blob-util/blob/80628d418a5c81439dc2c1a1f05fae6406f0ba7f/package.json#L46C10-L49))
+* Presence of React Native specific files such as `react-native.config.js` files
+Those indicators are all non-exhaustive.
```
lunaleaps, line 27: How do other libraries determine this? I guess it's not really a problem because we need to be able to catalogue libraries specifically for React Native?

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+## Rationale
+
+The need to specify **capabilities and metadata** for a react-native library is inherent to the React Native ecosystem.
+
+As of today, there is no simple programmatic way to know if a NPM package is a React Native library. Good indicators could be:
+* Presence of a `react-native:*` dependency in the `peerDependencies` section of the `package.json` ([example](https://github.com/RonRadtke/react-native-blob-util/blob/80628d418a5c81439dc2c1a1f05fae6406f0ba7f/package.json#L46C10-L49))
+* Presence of React Native specific files such as `react-native.config.js` files
+Those indicators are all non-exhaustive. 
+
+Similarly, there is no easy way to understand if a library is exposing a native module or not.
```
lunaleaps, line 29: What's the logic for needing to know whether a library exposes a native module or not?

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+  * The alternative at this stage is to inspect the code and check the API callsites if they are New Architecture or Old Architecture compatible.
+* Lack of a declarative way to enable New Architecture on both platforms
+  * Currently New Architecture is enabled by a Gradle Property `newArchEnabled` on Android in the `gradle.properties` file and by invokign `RCT_NEW_ARCH_ENABLED pod install` on iOS.
+  * As of today, there is no way to enable New Architecture for an app project for both platforms
+  * Moreover, the file vs environment variable leads to a scenario where you can't know if a app supports New Architecture or not by inspecting the code (as New Architecture support is known at  `pod install` time and is not codified in the codebase)
+
+Therefore we propose to add the `newArch` section to the `reactNativeManifest.capabilities` of **both apps and libraries** with the following semantic:
+* For Apps: `newArch.enabled==true` implies that the app wants to use the New Architecture.
+* For Libraries: `newArch.enabled==true` implies that the library is compatible with the New Architecture.
```
lunaleaps, line 72: Is the thinking then that there's no way to express one platform only supporting newArch? And it's not something we want to support?

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+* [ReactFeatureFlags.java](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/config/ReactFeatureFlags.java) which contains runtime feature toggles for Android.
+* [ReactNativeConfig.cpp](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/react/config/ReactNativeConfig.cpp) which contains runtime feature toggles for the C++ layer.
+
+Most of those config flags are undocumented and they're not verified for invalid configuration. Specifically, some feature flags require to be toggled on multiple layers, and it's easy to forget to toggle one of them. Invalid combinations of those flags could lead to unexpected runtime behavior.
+
+We propose to define `reactNativeManifest` as the **preferred entry point** for all the user facing entry points. All the tooling should be adapted to consume the information from the `reactNativeManifest` section and have sensible defaults defined by the tooling itself.
```
lunaleaps, line 189: Is this only for apps? I'm guessing it doesn't make sense for a library to specify flags? But I guess there could be a library that enforces a certain feature flag?

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+### JSON Schema
+
+The JSON Schema will be published as soon as we find an agreement on the key names and semantics.
```
blakef, line 302: Possibly publishing a validation library instead?  Something similar to [npmcli/package-json](https://www.npmjs.com/package/@npmcli/package-json) which users could use to parse, validate and evaluate the contents correctly (without having to home brew solutions).

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+
+It's outside the scope of this RFC to declare all the new sections we intend to add, and we defer to the [Future proof and extensibility](#future-proof-and-extensibility) section for the process on how to include a new key to the `reactNativeManifest` section.
+
+## Excluded use cases
+
+This section lists the use cases where we believe the `reactNativeManifest` section should **not** be used.
+
+### React Native version support
+
+For the time being, we're not planning to add a `version` section to the `reactNativeManifest` section for libraries. The rationale is that we believe that the `react-native` version should be specified in either the `peerDependencies` or the `engines` section of the `package.json` of the project. This follows the idiomatic way to specify the version of a dependency in a NodeJS project.
```
blakef, line 199: The examples below have a `version` section in the `reactNativeManifest`.  Is that a typo?

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+```
+
+For a library the section will look as follows:
+
+```json
+{
+  "name": "my-awesome-library",
+  "version": "1.2.3",
+  "reactNativeManifest": {
+    "version": "1.0",
```
blakef, line 278: Something that came up in discussion is what is the correct way to version and show semantic changes in a native module.  Currently the docs for TurboModules imply versioning your native module using the package.json's version for[iOS](https://reactnative.dev/docs/the-new-architecture/pillars-turbomodules#ios-create-the-podspec-file) (I'm not seeing anything like this for Android).  There's some work taking place to surface incompatibility between native modules built (dev mode) into the native app and their JS counterparts bundled by Metro.  

Examples would be:
- forgetting to build in a native module and giving clearer error messages; or
- changing branch which includes a significant library change and metro rebuilding your JS but the developer forgetting to rebuild your native app.

## Questions:
1. Should native extensions be using the `package.json → reactNativeManifest → version` or some combination of both the `package.json → version` and the `reactNativeManifest → version`?
2. Would we update the docs, file tickets with significant [new architecture Native Modules implementations](https://reactnative.directory/?ios=true&android=true&isMaintained=true&isPopular=true&newArchitecture=true)

blakef: See from the discussion with Nick that `reactNativeManifest → version` is a [schema version](https://github.com/react-native-community/discussions-and-proposals/pull/588#discussion_r1087184428).

proposals/0012-introduce-reactNativeMetadata.md
cortinico: Resolving as I believe this is not relevant anymore with the new proposal. Please read through again and share your concerns if any 👍

proposals/0012-introduce-reactNativeMetadata.md
cortinico: Yes, Metro config should live inside `metro.config.js`. Also this example has been removed from the proposal, so I'm resolving this comment

proposals/0012-introduce-reactNativeMetadata.md
cortinico: Resolving as this is now `capabilities`

proposals/0012-introduce-reactNativeMetadata.md
cortinico: Removing as this has been removed to reduce scope

proposals/0012-introduce-reactNativeMetadata.md
cortinico: Resolving as this is now `capabilities`

proposals/0012-introduce-reactNativeMeta.md
cortinico: We'll be having only `newArch` and not providing fine grain control. This is explained in the `### TurboModule/Fabric toggles` section I've added.

proposals/0012-introduce-reactNativeMetadata.md
cortinico: It's up to the various tools. The easiest would be to use https://www.npmjs.com/package/jsonschema inside the CLI and validate the `package.json` against the schema we provide.

cortinico: I've added a clarification about it in 68c906f36cef9af41a12172bc1c944900fb64f3b

proposals/0012-introduce-reactNativeMeta.md
cortinico: Clarified here: fe3e609f94490d9ac50213f5a16f71764c675d8a

> Could you give an example of the sort of breaking change you'd want to make at the schema level?

As long as we default of using objects all over, we'll be fine extending as we see fit. But I can see how a developer adds a `"mycapability": true` which then wants to extend with more metadata, and this is a breaking change.

proposals/0012-introduce-reactNativeMeta.md
cortinico: Thanks for pointing this out. I'm resolving this as `peerDependencies` got removed from this RFC as out of scope for now.

proposals/0012-introduce-reactNativeMeta.md
cortinico: > As this grows, or if we want to allow it to be dynamic, or extendable from a base, it seems beneficial to allow mirroring into a react-native.config.js outside the direct package.json 

Could you clarify on why we want this now? This is something we can still add later if we wish and the current design is not preventing it. Or am I missing something?

proposals/0012-introduce-reactNativeMeta.md
cortinico: Clarified on 76cb7db062beb14af3a610a87bce49785f51daa2

proposals/0012-introduce-reactNativeMetadata.md
cortinico: Clarified on 76cb7db062beb14af3a610a87bce49785f51daa2

proposals/0012-introduce-reactNativeMetadata.md
cortinico: > How do other libraries determine this?

What do you mean with "other libraries". Like other ecosystems?

proposals/0012-introduce-reactNativeMetadata.md
cortinico: > Why do we need to know whether a library exposes a native module or not?

For example: today we cannot crawl NPM to populate https://reactnative.directory/
Libraries needs to be added manually.

proposals/0012-introduce-reactNativeMetadata.md
cortinico: It's for both apps and libraries. There could be a scenarios where a capability needs all the dependencies to expose such capability (for example the `newArch` one).

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,329 @@
...
+    },
+    "ios": {
+      "capabilities": {
+        "newArch": {
+          "enabled": true
+        }
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
cortinico: Nope it's not. Top level capabilities applies to all platforms ("windows", "macos", others).

proposals/0012-introduce-reactNativeMetadata.md
cortinico: Yup also clarified here fe3e609f94490d9ac50213f5a16f71764c675d8a

cortinico:
> I want to push back against this point - the `peerDependencies` field brings a lot of baggage with it, and in the React Native ecosystem it often leads to a bad user experience in the form of unactionable/unhelpful error messages and obscure bugs related to package hoisting or duplicate versions of packages installed, where only a single version actually ever makes sense in the project (eg: for native modules, where only one version can be linked at any given time).

Thanks for flagging this @brentvatne. I think we'll probably have a separate RFC to decide how to proceed with either `peerDependencies`, `requirements` or any other mechanism to provide better support for version alignment. 

I do personally like your direction of allowing only `react-native` and `expo` as it will make it more predictable and easy to validate for users.

proposals/0012-introduce-reactNativeMetadata.md
lunaleaps: Oops yes other ecosystems. I'm guessing they have similar directories. Are they all maintaining it manually?

proposals/0012-introduce-reactNativeMetadata.md
lunaleaps: Would this proposal enforce this? Like if a library specifies a certain feature flag, and an app uses that library, how do we communicate this?

cortinico: > Would this proposal enforce this? Like if a library specifies a certain feature flag, and an app uses that library, how do we communicate this?

That's up to each feature to define the semantic of such flags. That's also why we need a dedicated RFC for each feature. There might be features that have a hard requirement on dependencies (like the New Architecture Renderer) and other features that have fallbacks

migueldaipre:
The `newArch` flag sounds strange to me. For me users new to the React Native ecosystem shouldn't know it as "newArch" or "oldArch".

cortinico:
> The `newArch` flag sounds strange to me. For me users new to the React Native ecosystem shouldn't know it as "newArch" or "oldArch".

For context, that's how it's called today:
https://reactnative.dev/docs/new-architecture-app-intro#android---prerequisites
Renaming features is outside the scope of this RFC

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+    "capabilities": {
+      "newArch": {
+        "enabled": true
+      }
+    }
+  }
+}
+```
+
+This section will allows also for split configuration between platforms:
```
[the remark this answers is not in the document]
cipolleschi: Shouldn't we specify the precedence rules? Because a setup like:

```json
{
  "reactNativeManifest": {
    "capabilities": {
      "android": {
      "capabilities": {
        "newArch": {
          "enabled": true
        }
      }
    }
  }
}
```

Is invalid. What is the definition for iOS?

Secondly: do we want to allow for Libraries to say that they supports the new architecture only for one of the two platforms or not?

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+#### `codegenConfig` support
+
+Similarly to New Architecture support metadata, the `codegenConfig` is a key metadata of the New Architecture build pipeline.
```
cipolleschi, line 115:
```suggestion
Similarly to the New Architecture support metadata, the `codegenConfig` is a key metadata of the New Architecture build pipeline.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+### Hermes Support
+
+Similarly to New Architecture support, the current way to enable/disable the Hermes engine is toggled by using a Gradle Property `hermesEnabled` on Android and changing the `:hermes_enabled` property in the Podfile.
```
cipolleschi, line 138:
```suggestion
Similarly to the New Architecture support, the current way to enable/disable the Hermes engine is toggled by using a Gradle Property `hermesEnabled` on Android and changing the `:hermes_enabled` property in the Podfile.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+    "capabilities": {
+      "hermes": {
+        "enabled": true
+      }
+    }
+  }
+}
+```
+
+which will also allow for split configuration between platforms as follows:
```
cipolleschi, line 154: Same consideration as above.

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+
+This section lists the use cases where we believe the `reactNativeManifest` section should **not** be used.
+
+### React Native version support
+
+For the time being, we're not planning to add a `version` section to the `reactNativeManifest` section for libraries. The rationale is that we believe that the `react-native` version should be specified in either the `peerDependencies` or the `engines` section of the `package.json` of the project. This follows the idiomatic way to specify the version of a dependency in a NodeJS project.
+
+### TurboModule/Fabric toggles
+
+At the time of writing, we prefer to don't offer a dedicated section to toggle Fabric/TurboModule capability inside `reactNativeManifest`. The rationale is that we believe that the `newArch` section is sufficient to expose the New Architecture to users.
```
cipolleschi, line 203:
```suggestion
At the time of writing, we prefer not to offer a dedicated section to toggle Fabric/TurboModule capability inside `reactNativeManifest`. The rationale is that we believe that the `newArch` section is sufficient to expose the New Architecture to users.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+Selectively toggling Fabric/TurboModule is a more advanced feature. We believe we'll still be offering a more advanced way to feature toggle such infrastructure pieces, but at the current state, our preference is to not expose such capability in the top level `reactNativeManifest` section.
```
cipolleschi, line 204:
```suggestion
Selectively toggling Fabric/TurboModule is a more advanced feature. We believe we'll still be offering a more advanced way to enable or disable those pieces of the New Architecture infrastructure, but at the current state, our preference is not to expose such capability in the top level `reactNativeManifest` section.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+
+1. If the capability `foo` is specificed in the `reactNativeManifest.capabilities` section, use the value specified in the `reactNativeManifest.capabilities.foo` section.
+   1. If the user is **also** specifying the capability `foo` in the build tool specific configuration (e.g. inside `gradle.properties` for Android) behave as follows:
+      1. If the two values are **compatible**, use them without notifying the user.
+      1. If the two values are **incompatible**, notify the user and use the value specified in the `reactNativeManifest.capabilities.foo` section will prevail and the value specified in the build tool specific configuration will be ignored.
+   1. If the user is **not** specifying the capability `foo` in the build tool specific configuration, honor the value specified in the `reactNativeManifest.capabilities.foo` section.
+1. If the capability `foo` is **not** specificed in the `reactNativeManifest.capabilities` section, honor the value specified in the build tool specific configuration or the default value if not specified.
```
cipolleschi, line 219: Can we be a little bit more aggressive?
So, for example, we can remove all the configuration flags in Cocoapods starting from version 0.73 (for example) so that the `reactNativeManifest` is the ONLY way to configure the app.
If there is no such file, we fail the build early, instructing the user how to set up those fields.

I know, it is a breaking change, but it will make the code much easier to write and follow and will let us converge to the ideal state faster. If the error messages are clear and actionable, most of the users will self-serve themselves.

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+
+### `align-deps` support
+
+This RFC originated from a [conversation between Meta and Microsoft](https://github.com/microsoft/rnx-kit/issues/1863) to use `align-deps` as key feature in the New Architecture rollout support, in helping developers understanding if a library they're using is New Architecture compatible or not. Very quickly it became clear that there is currently no straightforward way to know if a library indeed has that support.
```
cipolleschi, line 223:
```suggestion
This RFC originated from a [conversation between Meta and Microsoft](https://github.com/microsoft/rnx-kit/issues/1863) to use `align-deps` as key feature in the New Architecture rollout support, in helping developers understand if a library they're using is compatible with the New Architecture or not. Very quickly it became clear that there is currently no straightforward way to know if a library indeed has that support.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+
+Therefore we believe `align-deps` can benefit from this information and be extended to offer New Architecture support information.
```
cipolleschi, line 225:
```suggestion
Therefore we believe `align-deps` can benefit from this information and can be extended to offer New Architecture support information.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+
+### React Native CLI support
```
cipolleschi, line 227: would this be a part of React Native core or a feature we want to defer to frameworks?

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+
+We're versioning the schema using the `schema` key in the `reactNativeManifest` section. This allows us to evolve the schema in the future without breaking changes. We believe that the `schema` key should be a semver version string.
+
+We believe that the `reactNativeManifest` section should be extensible as we believe new capabilities and features can be added there.
+Specifically, we envision to evolve `reactNativeManifest` or `reactNativeManifest.capabilites` as follows:
+
+1. Each addition of a new capability [requires a new RFC](https://github.com/react-native-community/discussions-and-proposals#proposals) to be approved.
+2. Each RFC should define a new section in the `reactNativeManifest` with what's the intended use case.
+3. Each RFC should define what are the sensible defaults and should refrain from breaking changes.
+4. Each RFC should define the deprecation strategy for previous flags/features related to the same capability.
```
cipolleschi, line 250:
```suggestion
4. Each RFC should define the deprecation strategy for previous flags/features related to the same capability, if needed.
```

proposals/0012-introduce-reactNativeMetadata.md
```diff
@@ -0,0 +1,343 @@
...
+## Alternatives
+
+`reactNativeManifest` is, by design, fully original and new: there are other files that are used to define configurations for various aspect of react-native based projects (such as `react-native.config.js`, `app.json`, Expo's `app.config.js`, `expo-module.config.json`) but this one does **not** overlap with any of them. Explicitly, this one has the purpose of being filled only with capabilities and metadata to be read by package managers and build tools.
+
+## Adoption strategy
+
+If we decide to adopt this proposal, we'll have to broadcast this change to the ecosystem.
+Ideally, we'll start by adding the `reactNativeManifest` section to the app template and to `create-react-native-library`.
+
+We foresee a phase in which only a part of the userbase has this section configured and tool developers need to account for the absence of this section.
```
cipolleschi, line 327:
```suggestion
We foresee a phase in which only a part of the userbase has this section configured, and tool and developers need to account for the absence of this section.
```

cipolleschi (changes requested):
I left some comments. Most of them are fixes for typos, but I left also a couple of questions.

cipolleschi (changes requested):
I left some comments. Most of them are fixes for typos, but I left also a couple of questions.

cipolleschi:
Hey folks, I created [a(nother) PR](https://github.com/react-native-community/discussions-and-proposals/pull/717) with some small updates on this. 
Let's move the conversation there!

PR #607, opened 2023-03-01
RFC: DOM traversal and layout APIs in React Native

rubennorte:
This is a proposal to allow users to access a representation of the native view hierarchy in React Native using DOM-compatible APIs.

[A rendered version of the proposal can be read here](https://github.com/rubennorte/discussions-and-proposals/blob/dom-traversal-and-layout-apis/proposals/0607-dom-traversal-and-layout-apis.md)


necolas (approved):

tido64:
Hi, just some generic questions about this for my own understanding:
- Is this implemented in the JS layer only? Do we need to make any changes on the native side in core?
- What's the overhead of this API? Would it add to the final bundle size if unused?

rubennorte:
> Hi, just some generic questions about this for my own understanding:
> 
> * Is this implemented in the JS layer only? Do we need to make any changes on the native side in core?

@tido64 this also requires some new methods in the UIManager binding to expose the information to JavaScript. Something similar to these methods:

https://github.com/facebook/react-native/blob/851c8c671adcb563c9237e2080d5346413af1ac8/ReactCommon/react/renderer/uimanager/UIManagerBinding.cpp#L645-L677

We might also have to the a few methods in `UIManager` ([like this one](https://github.com/facebook/react-native/blob/851c8c671adcb563c9237e2080d5346413af1ac8/ReactCommon/react/renderer/uimanager/UIManager.cpp#L270-L304)). In any case, most of these are to access information in the shadow tree.

> * What's the overhead of this API? Would it add to the final bundle size if unused?

We expect this to be low and can optimize as needed. The implementation wouldn't be as large as it looks.

With some other APIs defined as globals we can optimize them by removing the global altogether if not used. With these methods it's going to be harder as they're exposed through refs. It's unlikely we'll be able to statically determine whether they're used or not.

sammy-SC (approved):

yungsters (approved):

javache (approved):

PR #637, opened 2023-04-20
RFC: Support Expo Config Plugins in React Native apps

TMisiukiewicz:
This RFC introduces the idea of upstreaming part of Expo Config Plugins into the React Native core, which can then be consumed by RN CLI, Expo, and others. We want Config Plugins to become a standard for building and configuring native parts of the apps across the React Native ecosystem.

Rendered RFC can be found [here](https://github.com/TMisiukiewicz/discussions-and-proposals/blob/rfc-config-plugins/proposals/0000-implement-config-plugins.md)

### DIFF
--- a/proposals/0000-implement-config-plugins.md
+++ b/proposals/0000-implement-config-plugins.md
@@ -280,6 +280,7 @@ This approach would also make it possible to extend config plugins with other pl
 
 ## Alternatives
 
+- installing Expo to the project and using `npx expo prebuild` - RN version is dependend on Expo SDK release cycle, however there is a `--template` flag which unlocks possibility of using any template with other version. This solution is not recommended by Expo because of making some undocumented assumptions about the template.
 - integrating `@expo/config-plugins` with React Native CLI to unlock a path for developers to run `yarn add @expo/config-plugins` in their non-Expo projects and use it - we might end up with a misalignment between what parts of Expo Config Plugins we want to support in the CLI (everything that is not related to Expo) and what developers will be able to use in their projects (everything, even if it's Expo-specific). This may lead to many misunderstandings why some of the properties in `app.json` or some mods are not working.
 - creating a new library based on Expo Config Plugins, exposing only stuff not related to Expo - it's not making Config Plugins a standard in the RN ecosystem, as it makes it totally optional to have it as a dependency in projects and libraries
 - creating a brand new CLI supporting Config Plugins - we won't fix the major problem of the apps currently running on React Native CLI - easy RN upgrades. It would be another tool existing in the RN ecosystem, still requiring to add a new dependency to projects/libraries and working only when using it with specific tool.
--- a/proposals/0000-implement-config-plugins.md
+++ b/proposals/0000-implement-config-plugins.md
@@ -1,0 +1,311 @@
+---
+title: Support Expo Config Plugins in React Native apps
+author:
+- Tomasz Misiukiewicz
+- Michał Pierzchała
+date: 20-04-2023
+---
+
+# RFC0000: Support Expo Config Plugins in React Native apps
+
+## Summary
+
+Upstream part of Expo Config Plugins into React Native core to support modifying application's config without having to manually edit the native files.
+
+## Basic example
+
+https://user-images.githubusercontent.com/13985840/229797752-81009e28-3153-4702-b52f-14602789a87b.mov
+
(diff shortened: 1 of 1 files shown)


pvinis:
I think that's a really good idea. for the longest time I've been thinking/trying to add xcodegen to RN for the app template, so that the upgrades can be simplified. i have not finished it yet, and I'm sad about that, but with this rfc I like the fact that this project generation can be common with expo. prebuild is in my opinion one of the best parts of expo, and the configs on top make it even better, do having those in RN too would be great. I'm all for this 👍.

pvinis (approved):

brentvatne:
i have lots of thoughts about this, but i think the highest order bit is: what is the issue with using the `expo` package and using `npx expo prebuild` if you want to use this functionality? ie: why does it need to be rebranded and forked, with effort in many ways duplicated, in order to avoid using that package and cli command? i could point to a bunch of other concerns i have with the proposal but i think before we get into any weeds here this is an important one for us to figure out.

- you can use it in projects created with npx react-native init, see: https://docs.expo.dev/guides/adopting-prebuild/
- support for out of tree platforms means opening up the `--platform` flag for prebuild, and otherwise building out support for it would be the same as it would be if this were part of the react-native community cli. we merged a PR that supports this in https://github.com/expo/expo/pull/22201 and we'll put together a small example to demonstrate how to use it

TMisiukiewicz:
Hi @brentvatne, thanks for your comment. Adding Expo to an existing project is an obvious alternative that I didn’t include in the RFC, thanks for pointing that out, will add it soon.

Expo prebuild works in majority of cases. However, it generates native code with RN version that depends on Expo release cycle. I think it is a very good solution, but we were thinking more of something that could go beyond that. I see a big potential in the `--template` flag which allows to work with different versions, played around with it a little bit and it seems like something that could be very beneficial in achieving that. However, as I can see in the Expo docs, it’s not generally recommended because the modifiers make some undocumented assumptions about the template files. I’m not very much into the technical details of it at this moment, but perhaps making a strict connection between any template and config plugins would reduce possibility of having any issues in this case. Are there any other drawbacks of using `--template` flag that you are aware of? I think it would also be potentially beneficial in terms of testing RC versions, as it would be easier to adopt by wider audience and collect more early feedback from users testing it in complex apps.

I’m really excited about opening up the `--platform` flag, seems to address out of tree platforms problem perfectly :+1: Great job on that one, we weren’t aware of that!

We’re also trying to discover how does this RFC relates to more advanced setups, like using alternative bundlers or brownfield development. I will try to keep everyone up to date with that once we have any outcome.

Our main goal here is to end up with a great upgrading experience for vast majority of RN users. We’re open to any kind of feedback, and even if our proposal turns out to be wrong, it’s a great opportunity to get all the parties involved and decide which way to go from there.

michalchudziak:
I love the proposal. It is a very convenient way to facilitate the update process, which is one of the most commonly reported pain points of working with the framework. Nevertheless, I would like to know about the hybrid/brownfield app perspective. If I understand it well, in such a case, you will most likely need to opt-out of using the plugin system/config generation. To be more specific, I'd like to find answers to the following questions:

- How would the process of integrating RN into existing native apps change?
- How easy would it be to add native screens directly to the app (without using the native module system)?
- What would happen if we run the prebuild command when the ios/android folders already exist? Will they be overridden? Is there any way to prevent that?
- Is there a convenient way to merge already existing manifest/plist file with `app.json` configuration?

TMisiukiewicz:
Hi @michalchudziak @blakef, thanks for your feedback. Happy the RFC brought some discussion around. 

I think the brownfield is an in interesting one. I can see some potential in utilizing already existing mods to automate some steps described in [Integration with Existing Apps](https://reactnative.dev/docs/integration-with-existing-apps) and when upgrading RN. The main problems I see here are:
- many times brownfield apps don’t follow the RN template structure which makes it harder to apply mods
- upgrading RN requires better understanding what changes are made in the RN template and how to correctly apply them into brownfield app
- Afaik there is no good way of parsing `.kts` files at this moment
- very common problem when upgrading brownfield apps, mostly on Android, are 3rd-party dependencies

If possible, we could try building some additional config around it to handle more complicated cases.  Additionally it could help developers easily handle any conflicts e.g. mismatching dependencies versions etc. Would love to hear more about problems the teams are facing with brownfield upgrade process to have a better understanding of this area

TMisiukiewicz:
@blakef since the #759 has already been merged, can we close this one? I think RFC for React Native Frameworks has set clear boundaries and it's result satisfies everyone's needs 👍

PR #646, opened 2023-05-02
RFC: JavaScriptCore-safe bundle URLs

robhogan:
Proposes a solution to the issues (https://github.com/facebook/react-native/issues/36794, https://github.com/expo/expo/issues/22008) caused by the [change to JavaScriptCore](https://github.com/WebKit/WebKit/blob/e214b2a322c49e1ff3aa7991b6f9478d8cbd80e5/Source/JavaScriptCore/runtime/StackFrame.cpp#L137) that strips information from stack trace source URLs.

### DIFF
--- a/proposals/0642-JavaScriptCore-safe-bundle-URLs.md
+++ b/proposals/0642-JavaScriptCore-safe-bundle-URLs.md
@@ -24,45 +24,41 @@ Should be supported by Metro (and other bundlers) equivalently and in addition t
 
 Recent versions of JavaScriptCore strip query strings from source URLs in stack traces (`Error.prototype.stack`).
 
-To implement features such as Metro's `/symbolicate` endpoint correctly, we must be able to derive (a close approximation of) the source that was executed the client. To rebuild that source correctly we need to use the same build parameters. 
+To implement features such as Metro's `/symbolicate` endpoint correctly, we must be able to derive (a close approximation of) the source that was executed the client. To do so correctly we need to know the build parameters the client used originally.
 
 Ideally, for maximal compatibility with debuggers (etc.) that may intercept the raw error stack, the source URLs in the error stack should be sufficient, and should be valid URLs, without any post-processing or external context.
 
 ## Detailed design
 
-Use a character sequence from the "reserved" set specified in RFC 3986, namely `;&`, as an alternative delimiter in place of `?`, and URL-encode any `?` characters appearing within the subsequent query string so that the result is a valid [`path`](https://www.rfc-editor.org/rfc/rfc3986#section-3.3).
+Use a character sequence from the "reserved" set specified in RFC 3986, namely `;&`, as an alternative delimiter in place of `?`, and URL-encode any `?` characters appearing within the subsequent query string, so that the result is a valid [`path`](https://www.rfc-editor.org/rfc/rfc3986#section-3.3).
 
 If the client sends a URL in this format, Metro should respond as if `?` was seen in place of the alternative delimiter, with the following exceptions:
  - Wherever the bundle itself encodes its own URL, source map URL, or the URLs of other bundles on the same server (such as for lazy bundling), those should use the alternative format for any included query strings *if and only if* the original bundle request used the alternative format. Otherwise, standard query strings starting with `?` should be emitted.
- - URLs of mixed format with `;&` preceding ?, i.e. `.../path;&foo=bar?foo=baz&bar=quux` should be interpreted leniently, as if they were given fully in the alternative format, i.e. `.../path;&foo=bar&foo=baz&bar=quux`.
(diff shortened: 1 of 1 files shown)


proposals/0642-JavaScriptCore-safe-bundle-URLs.md
```diff
@@ -0,0 +1,64 @@
...
+
+Because the alternative delimiter is made of reserved characters, it does not create ambiguity with standard URLs and may be escaped by percent-encoding for use as a param value, path component, etc. In particular, this supports the edge case of a valid POSIX file path containing the substring `;&` being the subject of a bundle request.
+
+## Drawbacks
+
+ - The proposed URL format is unfamiliar, and standard libraries and tools won't be able to encode/decode build parameters into/from URLs in this format. An exposed API or package would help mitigate.
+ - Maintenance/complexity of supporting an additional URL format.
+ - Forking of bundle URL generation in React Native by platform/engine, assuming we use this new format on as narrow a range of target engines as possible.
+
+## Alternatives
```
tido64, line 47: - What prevents Safari from also stripping out `;&` for security reasons?
- Have you considered embedding a comment similar to `//# sourceMappingURL=` for source maps?

robhogan: > - What prevents Safari from also stripping out ;& for security reasons?
 
Nothing in theory, but my understanding is that query strings (which of course aren't fundamentally less secure than path components) are commonly used to append tracking or other sensitive information in a web context during redirection. From what I can gather (though the security bugzilla is private), that was to plug quite a specific hole where stack consumers could end up being sent the sensitive data added by a separate redirection service. Those services would have to motivation to make that private data leakable, and couldn't append `;&` in the general case anyway because it will usually break routing.

I've also been in touch with Apple folks about building an opt-out for controlled, non-browser use cases like RN - they're open to that because the disruption to our use case was completely unintentional, so I don't think we're getting into an arms race here - just a workaround for a temporary situation.
 
> - Have you considered embedding a comment similar to //# sourceMappingURL= for source maps?

Metro already adds these, but JavaScriptCore doesn't use them when populating error stacks. The stacks coming from the engine only use the `sourceURL` provided explicitly to the `JSEvaluateScript` call ([docs](https://developer.apple.com/documentation/javascriptcore/1451589-jsevaluatescript)).

proposals/0642-JavaScriptCore-safe-bundle-URLs.md
```diff
@@ -0,0 +1,64 @@
...
+
+Ideally, for maximal compatibility with debuggers (etc.) that may intercept the raw error stack, the source URLs in the error stack should be sufficient, and should be valid URLs, without any post-processing or external context.
+
+## Detailed design
+
+Use a character sequence from the "reserved" set specified in RFC 3986, namely `;&`, as an alternative delimiter in place of `?`, and percent-encode any `?` characters appearing within the subsequent query string, so that the result is a valid [`path`](https://www.rfc-editor.org/rfc/rfc3986#section-3.3).
+
+If the client sends a URL in this format, Metro should respond as if `?` was seen in place of the alternative delimiter, with the following exceptions:
```
robhogan, line 35: @motiz88 I think we need to specify what "in this format" means here. For example, what format is this?
```
`/path/with-;&-unencoded/index.bundle?platform=ios`
```

If we interpret it as a "URL of mixed format" as suggested [below](https://github.com/react-native-community/discussions-and-proposals/pull/646/files#diff-0c487972bcf6173d648df22562629f0284700df0e5351731641cbf5ef18c41b6R37), then we'll break the specific edge case we claim to support, because clients are not presently required to percent-encode this sequence within a bundle request path (and in fact, don't). In practice, we *could* ensure that React Native escapes that sequence in future, but it'd still technically be a semver-breaking change to Metro, albeit in a narrow edge case.

It might be more practical to consider an input to be "in this format" if it contains *both* a `;&` within the path, *and no query component*.

Another possibility is to use a character sequence that can't appear in (normalised) POSIX paths - there aren't many, but maybe `//`?

robhogan:
I've just made a substantial update to the RFC in https://github.com/react-native-community/discussions-and-proposals/pull/646/commits/a96139248c84255d6d558e98b693a7a74baeb9c0, in summary:

 - The previous choice of delimiter `;&` did not allow us to disambiguate an encoded URL from an existing URL with `;&` coincidentally appearing in a file path, without assuming a client would percent-encode more of the path than it was required to. In particular, pre-existing clients would not be supported, which would make this change semver breaking. The choice of `//&` is explained in the RFC.
 - I'm now proposing that we adopt `//&` as the delimiter in *all* emitted "source URLs", and *not* requiring clients to update their request URLs in order to take advantage of the fix. Bundle request URLs are user-overridable in React Native, so our ability to control their formatting in a non-breaking patch is limited. By instead conceptually separating request URLs from source URLs, we can return a source URL along with the source from bundlers, and use that *downstream* of the http fetch.

PR #738, opened 2023-12-01
Add Support for Wide Gamut (DisplayP3) Colors to React Native

ryanlntn:
React Native does [not currently support](https://github.com/facebook/react-native/issues/41517) wide gamut color spaces (i.e. DisplayP3). This proposal discusses adding support for this color space in React Native.

Thanks to @cipolleschi for his support in drafting this proposal!


proposals/0000-add-support-for-display-p3-colors.md
```diff
@@ -0,0 +1,202 @@
...
+
+Since most new devices support wider gamut color spaces, React Native should support them as well. The DisplayP3 color space has had native support since Android 8.0 and iOS 9.3. The `color()` function was introduced with [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#color-function), much of which is already implemented in React Native. Flutter also [recently added support](https://github.com/flutter/flutter/issues/55092) for DisplayP3 on iOS with plans to follow up with support for Android and then the framework itself. Support for DisplayP3 colors can improve the quality of React Native apps by providing developers a wider range of available colors to use and is aligned with React Native's mission to provide the same look and feel as the underlying native platforms.
+
+## Detailed design
+
+### JS Changes
+
+1. Parse the color() function in [normalizeColor](https://github.com/facebook/react-native/blob/63213712125795ac082597dad2716258b90cdcd5/packages/normalize-color/index.js#L235)
+
+```js
```
NickGerleman, line 51: The motivation and API here make sense. How close is this towards a PR?

NickGerleman:
Also FYI @vincentriemer as someone who has been interested in extended color space before.

proposals/0000-add-support-for-display-p3-colors.md
ryanlntn: I haven't done any implementation work yet. Riccardo has been helping me with this so we were planning on leaving this open for feedback for a couple weeks before starting on implementation when he gets back.

Kudo:
cc @andreialecu who is also interested in this topic.

my understanding is that, unlike flutter to draw everything by their own. by leveraging the UIImage, we have the display p3 support already. the only missing piece is our coloring system as this pr pointed out.

alternatively, i was thinking to change from native code only without touching js code. if the screen supports display p3, we should just use the display p3 UIColor methods. having a small poc here: https://github.com/Kudo/displayp3/blob/main/modules/p3color/ios/p3colorView.swift
we can check whether the view support display p3 by `self.traitCollection.displayGamut`. then we can get this information from RCTRootView and pass down to some new UIColor factory.

proposals/0000-add-support-for-display-p3-colors.md
```diff
@@ -0,0 +1,202 @@
...
+
+2. **Utilizing Swizzling on iOS:** <br/> As an alternative to the proposed iOS changes, swizzling could be considered similar to [RN #41517](https://github.com/facebook/react-native/issues/41517). This has a major downside of being an all or nothing breaking change forcing all React Native users into the DisplayP3 color space as well as being a more confusing and less maintainable solution.
+
+3. **Full switch to DisplayP3:** <br/> In a similar vein we could just update the constructors directly without preserving existing behavior. While this would be a simpler implementation it would be a breaking change. Any existing visual regression tests would likely break and React Native users would now be required to specify all colors in a wider gamut.
```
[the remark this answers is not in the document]
ryanlntn: > What do you think about a way to optionally fully opt-in to use DisplayP3?

I can definitely see the utility of this. I'm not really sure what the API for that should look like though. AFAIK the APIs on Android, iOS, and web all only allow specifying color space per instance. Since color properties already mostly implement CSS Color Module Level 4 we were mainly planning on extending that further with this proposed API.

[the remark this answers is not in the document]
cipolleschi: I agree that it could be useful. 
The way I think this should be implemented is by factoring out the constructor functions in a single factory method so we can decide which initializer we want to use as default. At that point, we can specify a static feature flag that controls the default behavior.

For Android things are a little bit more convoluted because Android uses `int` to represent sRGB colors and `long` to represent WideGamut colors. So, not only we need to update how a color is created but we also have to overload several methods to achieve what we need.

But it is probably doable.

@ryanlntn could you update the proposal to add a feature flag that will control the default init behavior?

proposals/0000-add-support-for-display-p3-colors.md
```diff
@@ -0,0 +1,254 @@
...
+### Using a feature flag
+
+Opt-in to using DisplayP3 as the default color space by using a feature flag.
+
+```sh
+# in ios/Podfile
+ENV['RCT_WIDE_GAMUT_ENABLED'] = '1'
+
+# in android/gradle.properties
+wideGamutEnabled=true
```
cipolleschi, line 50: I'd rather not use this build-time approach as it pollutes the codebase with checks all-over the place.

We can use feature flags like this one in the [`RCTBridge`](https://github.com/facebook/react-native/blob/main/packages/react-native/React/Base/RCTBridge.h#L45-L46).

In our case, they will not go in the RCTBridge file, of course, but this is a detail we can figure out later.

How it expect to work is that we put the creation of a color in its own factory function that takes the color parameters and the color space.
If the color space is not passed, it reads from the global feature flag what is the default space.
But we should try to avoid compile time flags as they are hard to read and they are hard to maintain.

proposals/0000-add-support-for-display-p3-colors.md
```diff
@@ -0,0 +1,254 @@
...
+struct ColorComponents {
+  float red{0};
+  float green{0};
+  float blue{0};
+  float alpha{0};
+#if RCT_WIDE_GAMUT_ENABLED
+  ColorSpace colorSpace{ColorSpace::DisplayP3};  // Default to DisplayP3
+#else
+  ColorSpace colorSpace{ColorSpace::sRGB};  // Default to sRGB
+#endif
```
cipolleschi, line 160: we should avoid this compiler flag

cipolleschi (changes requested):
Can we try to avoid compile time flags? 🙏

henrymoulton:
A tweet around this went semi-viral https://twitter.com/lauridskern/status/1717237553886081252

eliotgevers:
Any update?

Issue #253, opened 2020-06-24
Transition to Swift


yvbeek:
### Introduction

When development on React-Native started, Objective-C was the primary languages for all iOS apps. On June 2, 2014, a little under a year before the first public release of React-Native, Apple's new programming language Swift was introduced to the world.

The first few versions of Swift were not ready for production apps. New releases often introduced code breaking changes and I think we all have "fond" memories of the Xcode SourceKit crashes and other stability issues.

Fast forward to March 2019, Swift 5 was introduced with ABI stability. Many developers have now adopted the language and Apple is pushing the language forward with a new Swift-only UI framework called SwiftUI. During WWDC 2020 we learn that parts of MacOS Big Sur (for example the new control center) are implemented with SwiftUI.

Clearly Objective-C is moving into the background and the path forward is Swift. My proposal for React-Native would be to start moving into the same direction. 

### Details

When you create a new React-Native project, you are greeted by an Objective-C project. With some work you can convert the `AppDelegate` and other classes to Swift and you'll have to figure out how to get the Podfile properly configured for Swift <=> Objective-C interoperability.

This isn't a great experience for new developers starting with React-Native, nor for developers that already have a native app written in Swift and that want to transition to a multi-platform React-Native app.

When it comes to device support, the first iOS version able to run Swift apps was iOS 7. The minimum requirement for React-Native nowadays is iOS 10, therefor Swift support is not a problem.

But why Swift? I think there could be a lot to gain from the safety features that are built right into the Swift language. Code written in Swift is often shorter and uses a modern syntax that reduces code complexity.

### Discussion points

- What are the specific Objective-C features that React-Native is using (things like macros)?
- What are up- and downsides of _slowly_ transitioning the codebase to Swift?
- What would be the best steps to move towards a Swift codebase (new files in Swift, patch code with extensions)?
- How can we make React-Native more accessible to Swift developers?
- How can we make it easier to create React-Native community packages in Swift?

(we could start a similar proposal for Android for Kotlin)

yvbeek:
Linking to [What don't you like about React-Native](https://github.com/react-native-community/discussions-and-proposals/issues/134#) => [Swift not being the default language for iOS](https://github.com/react-native-community/discussions-and-proposals/issues/134#issuecomment-501279947)
The comment is from June 2019 and at the time of writing has 212 upvotes

mrousavy:
For the record, in many cases Objective-C is still "faster" than Swift.

[...]

PR #747, opened 2023-12-14
Adding support for Swift / Moving away from Obj-C on Apple platforms

TheRogue76:
This proposal aims to define a possible design for how we can move away from Obj-C on Apple platforms, and what challenges we might face along the way and what needs to be done to get there. It is a continuation of the work setup by https://github.com/react-native-community/discussions-and-proposals/issues/253 and https://github.com/facebook/react-native/pull/41896 .

Please feel free to leave comments, edits and thoughts on this proposal as well as any challenges that you feel might block  such a change so we can define the scope of the work that might be needed and how / if we should go about doing this.

Special thanks to  @cipolleschi for the instructions on how to setup this RFC

### DIFF
--- a/proposals/0000-adding-support-for-swift.md
+++ b/proposals/0000-adding-support-for-swift.md
@@ -21,6 +21,54 @@ The motivation for this change is threefold:
 
 ## Detailed design
 
+### TODO: I am unaware of a lot of the internals of RN, so if anyone more aware than me can contribute, it is appreciated
+
+The general idea is that the code we have for Apple platforms should be consistent of three primary parts:
+- The Swift API layer that is in charge of the platform specific operations that can not be shared across other platforms (the View Delegate, the Scene Delegate, anything that has to do with Apple specific frameworks like UIKit)
+- The Shared C++ code that is platform agnostic and used by all React Native target platforms (Yoga, Hermes, the event handlers, the bridge (if applicable), etc)
+- The optional intermidery C++ code, with Swift specific annotations for certain APIs that help Swift interact with C++ in a more efficient and correct manner (To learn more, please refer to [Swift's documentation of interop with C++](https://www.swift.org/documentation/cxx-interop/#exposing-swift-apis-to-c) under "Customizing How C++ Maps to Swift")
+
+The third step is marked as optional. The idea is that the Swift compiler makes certain assumptions when interacting directly with the C++ modules that may not necessarily be correct (C++ types being imported as value types by default, for instance). But we can override those behaviours by explicitly marking certain sections with the annotations provided to us by <swift/bridging> (Marking things to be imported as reference types, Mapping getters and setters, etc)
+
+Much like our current setup with Obj-C, The Apple platform section of any new project template created will simply expose the Swift files for developers to extend and add functionalities to. The React Native APIs can be simply imported via a swift import and consumed by the Swift parts, giving us control over our API surface and hiding away any specific pieces that will not be needed on a day to day basis. Example:
+
+```swift
+import React
+
+@main
+class AppDelegate: RCTAppDelegate {
+    override func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        self.moduleName = "HelloWorld"
+        // You can add your custom initial props in the dictionary below.
+        // They will be passed down to the ViewController used by React Native.
+        self.initialProps = [:]
+
+        return super.application(application, didFinishLaunchingWithOptions: launchOptions)
+    }
+
+    override func sourceURL(for bridge: RCTBridge!) -> URL! {
+        return self.getBundleURL()
+    }
+
+    func getBundleURL() -> URL {
+        #if DEBUG
+        return RCTBundleURLProvider.sharedSettings().jsBundleURL(forBundleRoot: "index")
+        #else
+        return Bundle.main.url(forResource: "main", withExtension: "jsbundle")!
+        #endif
+    }
+
+    /// Native bits of functionality that users may extend and add on their own
+}
+```
+
+
+
+
+
+
+
+
 This is the bulk of the RFC. Explain the design in enough detail for somebody familiar with React Native to understand, and for somebody familiar with the implementation to implement. This should get into specifics and corner-cases, and include examples of how the feature is used. Any new terminology should be defined here.
 
 ## Drawbacks
--- a/proposals/0000-adding-support-for-swift.md
+++ b/proposals/0000-adding-support-for-swift.md
@@ -21,14 +21,14 @@ The motivation for this change is threefold:
 
 ## Detailed design
 
-### TODO: I am unaware of a lot of the internals of RN, so if anyone more aware than me can contribute, it is appreciated
+### TODO: I am unaware of a lot of the internals of RN, so if anyone more aware than me can contribute, it is appreciated.
 
-The general idea is that the code we have for Apple platforms should be consistent of three primary parts:
+The general idea is that the code we have for Apple platforms should be consistent of three primary parts by the end of this migration:
 - The Swift API layer that is in charge of the platform specific operations that can not be shared across other platforms (the View Delegate, the Scene Delegate, anything that has to do with Apple specific frameworks like UIKit)
 - The Shared C++ code that is platform agnostic and used by all React Native target platforms (Yoga, Hermes, the event handlers, the bridge (if applicable), etc)
 - The optional intermidery C++ code, with Swift specific annotations for certain APIs that help Swift interact with C++ in a more efficient and correct manner (To learn more, please refer to [Swift's documentation of interop with C++](https://www.swift.org/documentation/cxx-interop/#exposing-swift-apis-to-c) under "Customizing How C++ Maps to Swift")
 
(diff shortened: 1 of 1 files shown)


proposals/0000-adding-support-for-swift.md
```diff
@@ -0,0 +1,111 @@
...
+
+The biggest drawback can be boiled down to certain key elements:
+- It is a massive undertaking, that will likely take years to fully complete
+- The Swift API is not yet fully compatible with everything we will need on the C++ side of things
+- Libraries will need to update their documentation and probably some changes to their imports as well (ideally, the API should not change, unless it is related to the Platform specific side of things and the Swift implementation simplifies the existing API for them to consume) and as [previously established](https://github.com/react-native-community/discussions-and-proposals/issues/671) not all libraries adapt and maintainers may not update
+- Users will need to upgrade their templates and the libraries they use to Switch over
+
+## Alternatives
+
+### Leaving things as is
```
cipolleschi, line 75: There is another alternative: we can create a wrapper of Objective-C(++) around C++ pure code exposing APIs that are compatible with Swift.
This approach is a middle-ground that should:
- avoid a complete rewrite of some internals
- allow contributors to start using swift only
- work around current Swift/C++ interop limitations
- take much less time than the main approach

To achieve this, we should highlight which are all the APIs that could be useful for devs to be accessed in Swift (which is not a joke of a task)

proposals/0000-adding-support-for-swift.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Unresolved questions
+
+- What other parts of the infra need to be changed for this transition to work?
+- How will this affect other targets such as React Native macOS, React Native visionOS, etc.?
+- Is our current separation of internals from the OSS template enough that we can simply change the template directly without any of the first four steps needed? (I managed to get a [test repo](https://github.com/TheRogue76/React-Native-Template-With-Swift) to work as a proof of concept with these changes, but I am not sure if that means that the app will work for all situations (new arch, bridge-less, etc.). I also submitted [a PR](https://github.com/facebook/react-native/pull/41896) to test out the current tests we had and those passed, but that might not be enough). And what about Expo? Will it continue to function if this change is done to the base template, or would it interfere with any of the work done on their side
```
cipolleschi, line 109: > Is our current separation of internals from the OSS template enough that we can simply change the template directly without any of the first four steps needed?

No, as explained in the PR, that's not enough.
We had to make sure that all the configurations work and that the flags are passed properly. 

Also, the AppDelegate is not the only API that users interface with. There are all the APIs related to create a Custom Module and a Custom Component.

proposals/0000-adding-support-for-swift.md
```diff
@@ -0,0 +1,111 @@
...
+
+This should be a pretty welcome step for the user base (and even the maintainers, if done correctly and without too many breaking changes). The docs for creating Native module will need to be updated to add support for Swift, and we can simply mention the change in React Native EU or similar conferences.
+
+The idea is to make it very un-noticeable to the vast majority of the user base who simply want to use the React side for creating their mobile applications, and the part of the community that is capable of creating native modules should simply feel like it is just another native Apple app that they can extend using whatever it is that they need (perhaps adding a watchOS target support for their app, or using CoreML or AppClips or making a SwiftUI based widget)
+
+It should be presented as React Native keeping up with changes and lowering the barrier of entry for all devs and that we are simply making part of the work that is duplicated across platforms easy to reuse, using the same tools that our industry is using and without sacrificing the experience.
+
+## Unresolved questions
+
+- What other parts of the infra need to be changed for this transition to work?
```
cipolleschi, line 107: Surely, we need to verify that
- Native/TurboModule APIs are compatible with Swift
- Native/Fabric Components APIs are compatible with Swift
- Codegen (for the New Arch) is compatible with Swift (and can generate Swift code)

proposals/0000-adding-support-for-swift.md
```diff
@@ -0,0 +1,111 @@
...
+- Updating React Native Helper to support people trying to migrate their templates from Obj-C to Swift
+- Updating the docs
+- Communicating with library maintainers if help is needed updating their documentation to support the setup for both Swift and Obj-C for older users
+- Rolling out a new version
+- Moving as much shared logic as possible to C++ and away from Obj-C, separating them as distinctly as possible while adding new features and capabilities and continuing to roll out new versions (**Users should not be affected, and ideally the Library maintainers should not be affected either**)
+- Beginning the transition from Obj-C to Swift from the tests in the internals (**No effect on users or library maintainers**)
+- Moving the rest of the modules one at a time from Obj-C to Swift (**Library maintainers may need to update, if they wish to take advantage of the newer APIs or if we break change an APIs signature, but ideally it should only be in the direction of making it easier to maintain libraries and better APIs**)
+- Updating CocoaPods infra
+- ? (Please add any more steps that come to mind, at the end or anywhere else in the list)
```
cipolleschi, line 95: There is an important step missing.
The New Architecture main language is C++ and it is all conatined in the `ReactCommon` folder.
Most of the code exposed from `ReactCommon` is not configured as a Clang Module. This would make impossible for that code to be consumed by Swift, as [documented here](https://www.swift.org/documentation/cxx-interop/#creating-a-clang-module)

cipolleschi (commented):
Hi @TheRogue76 Thanks for opening this RFC. I left some comments with more context.

To set expectations, our plans for next year are to double down on the New Architecture work. The final goal is to remove the old architecture (we don't have an estimate yet of when this will happen) but we will not have capacity to work on Swift support unfortunately.

As soon as the New Arch is out of the door and the old arch is deprecated, I hope we will have more bandwidth. It will also be easier as we won't have to support both architecture but we will be able to only focus on the New one.

proposals/0000-adding-support-for-swift.md
```diff
@@ -0,0 +1,111 @@
...
+- Obj-C is relatively hard to write and maintain for, and it is reflected by the [community](https://github.com/react-native-community/discussions-and-proposals/issues/104). Having a modern language that is more readable, more performant(in some cases) and [safer from certain class of errors](https://developer.apple.com/swift/#safety) means that users and library maintainers alike can produce better quality code and take advantage of the underlying platforms easier if they need it
+- As Swift evolves and becomes more feature rich, there is a possibility that Apple will begin to remove support for Obj-C in it's toolchain. The documentation for the various APIs that Apple provides may also be moved to simply having Swift as the supported language.
```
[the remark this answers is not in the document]
TheRogue76: Very true. Added a line to it under [21e7c3f](https://github.com/react-native-community/discussions-and-proposals/pull/747/commits/21e7c3f63aa0213a214744a31b12d5b0b681cef1)

proposals/0000-adding-support-for-swift.md
TheRogue76: Added under [d5c39ae](https://github.com/react-native-community/discussions-and-proposals/pull/747/commits/d5c39aea9f2196cbdcbf3f2a44d22d2baf149491). Could be a nice alternative. Wrote one con for it as well that comes to my mind regarding Toolchain depracation but perhaps it won't be that big of a problem

proposals/0000-adding-support-for-swift.md
TheRogue76: Removed under [6a05439](https://github.com/react-native-community/discussions-and-proposals/pull/747/commits/6a0543926e0f59ec507fe97e4cd97834aaa7efde).

proposals/0000-adding-support-for-swift.md
TheRogue76: Mentioned these as questions that need to be solved under [8ecb9db](https://github.com/react-native-community/discussions-and-proposals/pull/747/commits/8ecb9db79244b93ded0bebf0276d33f6115cc460).

proposals/0000-adding-support-for-swift.md
TheRogue76: Ah, that explains quite a bit. So would we need to configure them as Clang Modules? Is it even possible or would it have some other consequences on other parts of the code base?

cipolleschi: It is for sure possible.
It will require to write some modulemap files. If that's all it takes, it would not be breaking.
But my gut feeling tells me that we will have to also move some files around, specifically we should move the public headers to an `include` folder. And that would be breaking, unfortunately.

TheRogue76: Mentioned both the required action and the concern as steps we would need to take under [228b56d](https://github.com/react-native-community/discussions-and-proposals/pull/747/commits/228b56dbbd63f16a1da25d2122fabd32207c2e91)

mrousavy:
Not very constructive, but I just wanna say I really stand behind this RFC.

Objective-C (or actually Objective-C++) is really good when it comes to C++ interop, but that's about it. Swift is easier to read, much less code, simpler, memory is managed better, sometimes it's even faster (direct method call vs message send), and it ultimately invites more outside contributions.
I'm sure a lot of people are hesitant about reading RN Core code or contributing to it just because it's Objective-C. Swift is easier to understand for JS devs.

Expo uses Swift, [VisionCamera](https://github.com/mrousavy/react-native-vision-camera) uses Swift, and pretty much all modern iOS libs use Swift.

With that said, it's obviously a huge effort, and I fully agree that current efforts are best dedicated to new arch stuff.  That's where it starts to get tricky though because everything is in C++ now and Swift <> C++ interop is still very limited - no templates, not all default headers will work, and stuff like folly, jsi, or react headers will probably require intermediate bridging headers.

Which is an _okay_ workaround, but takes more time to write.

PR #750, opened 2024-01-08
Cursor pointer API 

okwasniewski:
This PR describes a new `cursor: "pointer"` API for iPad and Apple Vision Pro.

I've created this proposal together with @SaadNajmi

# [A rendered version of the proposal can be read here](https://github.com/okwasniewski/discussions-and-proposals/blob/hover-style-api/proposals/0000-hover-style-ipad-api.md)

### DIFF
--- a/proposals/0000-hover-style-ipad-api.md
+++ b/proposals/0000-hover-style-ipad-api.md
@@ -1,0 +1,104 @@
+---
+title: Hover style API for iPad / Apple Vision Pro
+author:
+  - Oskar Kwaśniewski
+  - Saad Najmi
+date: 08.01.2023
+---
+
+# RFC0000: Hover style API for iPad / Apple Vision Pro
+
+## Summary
+
+iPad supports hover style effects when using an external mouse/trackpad. This effect illuminates the UIView in two configurable ways: lift or highlight. This proposal intends to add support for this API in React Native. This API is also used by visionOS to display the hover effect when a user looks at the view.
+
+The native API this feature is basing on: https://developer.apple.com/documentation/uikit/uihoverstyle (iOS 17+). There is an older version of this API supporting iOS 13.4+ https://developer.apple.com/documentation/uikit/uipointerstyle
+
+## Basic example
+
+This API would introduce an additional prop to UIView (which is used to render touchable/pressable elements).
+
+```jsx
+<Pressable hoverEffect=”lift” onPress={onPressFunction}>
+  <Text>I'm pressable!</Text>
+</Pressable>
+```
+
+The hover effect automatically adapts to view’s corner radius.
+
+Example of highlight effect:
+
+https://github.com/react-native-community/discussions-and-proposals/assets/52801365/a1ace4ac-44c1-45e6-afbe-6e0b0f05f74d
+
+## Motivation
+
+Controlling hover style allows React Native developers to provide a much better user experience on iPad and Apple Vision Pro, allowing them to seamlessly interact with React Native applications using the mouse.
+
+Additionally, We have lots of React Native experiences in brownfield contexts at Microsoft. The value add of React Native over something like a webview is that we can access the system APIs, and provide the best UX for the platform. This includes iPadOS / visionOS, where hover events are now a part of the UX language and built into UIKit controls. In order for React Native experiences to thrive living side by side with native experiences, we would live to have as close UX as possible.
+
+This API can work side by side with the Pointer Events API. Users can choose to either spin up their own custom interactions with the pointer events API, or use Apple OS default. Adding a hover effect shouldn’t affect pointer events being fired or not. https://reactnative.dev/blog/2022/12/13/pointer-events-in-react-native
+
+## Detailed design
+
+_Disclaimer: I will discuss the design for New Architecture, but the API can be easily backported._
+
+Also this API is already working in react-native-visionos [fork](https://github.com/callstack/react-native-visionos/blob/58f6e180d49cbd638ff9b637c0308808eeee2321/packages/react-native/React/Fabric/Mounting/ComponentViews/View/RCTViewComponentView.mm#L522).
+
+The API should be opt-in to avoid breaking current iPad apps. The implementation bases on extending `RCTView` with a new prop of type `std::string`.
+
+Next inside of `RCTViewComponentView` `- (void)updateProps:(const Props::Shared &)props oldProps:(const Props::Shared &)oldProps`` method:
+
+```objc
+if (oldViewProps.hoverEffect != newViewProps.hoverEffect) {
+    [self updateHoverEffect:[NSString stringWithUTF8String:newViewProps.hoverEffect.c_str()]];
+  }
+
+In the code above we are detecting props update, and calling updateHoverEffect method:
+
+- (void) updateHoverEffect:(NSString*)hoverEffect {
+    if (hoverEffect == nil || [hoverEffect isEqualToString:@""]) {
+        self.hoverStyle = nil;
+        return;
+    }
+
+    UIShape *shape = [UIShape rectShapeWithCornerRadius:self.layer.cornerRadius];
+    id<UIHoverEffect> effect;
+
+    if ([hoverEffect isEqualToString:@"lift"]) {
+        effect = [UIHoverLiftEffect effect];
+    } else if ([hoverEffect isEqualToString:@"highlight"]) {
+        effect = [UIHoverHighlightEffect effect];
+    }
+
+    if (hoverEffect != nil) {
+        self.hoverStyle = [UIHoverStyle styleWithEffect:effect shape:shape];
+    }
+}
+```
+
+Here we initialize a new `UIShape` which reassembles the current’s view layer `cornerRadius`. Next we are creating a `UIHoverEffect` instance based on the value passed by user.
+
+The current API implementation might need some polish to correctly retrieve corner radius.
+
+## Drawbacks
+
+- Apple only API
+  - Only Android equivalent I found is: https://developer.android.com/reference/android/view/MotionEvent which doesn’t provide any built-in effect but allows to create custom ones.
+- There might be different plan for the Pointer Events API.
+
+## Alternatives
+
+- This API might be also implemented using Pointer Events API by detecting hover and applying some custom styling to the view but this would be less performant than using native solution.
+- This feature could be also implemented in user space by creating their own subclass of UIView and wrapping every element with it.
+
+## Adoption strategy
+
+Users can opt-in without any breaking changes.
+
+## How we teach this
+
+Describe this prop in the documentation. Add section to docs about iPad development.
+
+## Unresolved questions
+
+- How does it fit with Pointer Events API?
--- a/proposals/0000-hover-style-ipad-api.md
+++ b/proposals/0000-hover-style-ipad-api.md
@@ -6,7 +6,7 @@ author:
 date: 08.01.2023
 ---
 
-# RFC0000: Hover style API for iPad / Apple Vision Pro
+# RFC0000: Hover style API for iPad / Apple Vision Pro (Cursor: pointer)
 
 ## Summary
 
@@ -16,10 +16,10 @@ The native API this feature is basing on: https://developer.apple.com/documentat
 
 ## Basic example
 
-This API would introduce an additional prop to UIView (which is used to render touchable/pressable elements).
+This effect would be applied by adding `cursor: pointer` style to a view (which is used to render touchable/pressable elements).
 
 ```jsx
-<Pressable hoverEffect=”lift” onPress={onPressFunction}>
+<Pressable style={{ cursor: "pointer" }} onPress={onPressFunction}>
   <Text>I'm pressable!</Text>
 </Pressable>
 ```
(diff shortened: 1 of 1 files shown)


proposals/0000-hover-style-ipad-api.md
```diff
@@ -0,0 +1,104 @@
...
+
+https://github.com/react-native-community/discussions-and-proposals/assets/52801365/a1ace4ac-44c1-45e6-afbe-6e0b0f05f74d
+
+## Motivation
+
+Controlling hover style allows React Native developers to provide a much better user experience on iPad and Apple Vision Pro, allowing them to seamlessly interact with React Native applications using the mouse.
+
+Additionally, We have lots of React Native experiences in brownfield contexts at Microsoft. The value add of React Native over something like a webview is that we can access the system APIs, and provide the best UX for the platform. This includes iPadOS / visionOS, where hover events are now a part of the UX language and built into UIKit controls. In order for React Native experiences to thrive living side by side with native experiences, we would live to have as close UX as possible.
+
+This API can work side by side with the Pointer Events API. Users can choose to either spin up their own custom interactions with the pointer events API, or use Apple OS default. Adding a hover effect shouldn’t affect pointer events being fired or not. https://reactnative.dev/blog/2022/12/13/pointer-events-in-react-native
```
NickGerleman, line 39: @javache shared that Vision Pro SDK does not currently allow tracking gaze, beyond declarative hover events, to protect privacy of what the user is looking at. Under that model, pointer events don't seem possible.

proposals/0000-hover-style-ipad-api.md
```diff
@@ -0,0 +1,104 @@
...
+The native API this feature is basing on: https://developer.apple.com/documentation/uikit/uihoverstyle (iOS 17+). There is an older version of this API supporting iOS 13.4+ https://developer.apple.com/documentation/uikit/uipointerstyle
+
+## Basic example
+
+This API would introduce an additional prop to UIView (which is used to render touchable/pressable elements).
+
+```jsx
+<Pressable hoverEffect=”lift” onPress={onPressFunction}>
+  <Text>I'm pressable!</Text>
+</Pressable>
```
NickGerleman, line 24: Having unprefixed, single platform props for shared components like `Pressable` is a little bit smelly.

There are various schemes for prefixing and suffixing platform specific props in RN, and also ones we can do that we haven't tried yet (`<Pressable apple:hoverEffect="lift">`)

This might actually belong on the Platform's View (or `VisionOSView` instead), if we would ever want to support hover style outside Pressables.

proposals/0000-hover-style-ipad-api.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Drawbacks
+
+- Apple only API
+  - Only Android equivalent I found is: https://developer.android.com/reference/android/view/MotionEvent which doesn’t provide any built-in effect but allows to create custom ones.
+- There might be different plan for the Pointer Events API.
+
+## Alternatives
+
+- This API might be also implemented using Pointer Events API by detecting hover and applying some custom styling to the view but this would be less performant than using native solution.
```
NickGerleman, line 91: RN on Quest ends up doing something like:
1. Set declarative information on view
2. In native, on pointer move, show underlay, vibrate controllers, etc on native hit test, before ever dispatching to JS

Some of this is exposed via `Animated.event`, which is a facility to animate native props when firing en event, before delivered to JS.

proposals/0000-hover-style-ipad-api.md
Saadnajmi: To clarify, are you saying the existing pointer events API in React Native won't work in visionOS? The new hover effects that visionOS added seemed to me something separate from the existing APIs, which I thought should still work without issue?

proposals/0000-hover-style-ipad-api.md
Saadnajmi: In my head, something like `hoverEffectIOS` o hoverEffectApple` is fine, though I feel it has been somewhat inconsistent in the past ( `elevation` isn't `elevationAndroid`, etc..). As implemented in the visionOS fork right now, it is a prop on the base `<View>`, and `<Pressable>` just sets a default that isn't `none`.

vincentriemer:
Thank you so much for putting together this RFC! I’m more or less going to take this opportunity to elaborate on my own alternative to this proposal that I’ve been thinking about for a while now, but didn’t have the bandwidth/priority to put one together myself.

The summary/motivation you’ve described in this RFC is extremely similar to what I was thinking but there are big differences in how we approach the API design. As additional context I am part of the XPlat React/Convergence workstreams focusing on interoperability between platforms including web, and that greatly informs the lens I look through when approaching these APIs.

As the person who is working on implementing Pointer Events, I was immediately curious as to the Vision Pro could take advantage of my work when it was announced. I quickly learned that the underlying eye gaze coordinates are never provided raw to userland code (unless an indirect/direct gesture is underway). This is instead done declaratively either through the APIs mentioned in the parent RFC or by default with Apple’s provided UI elements like UIButton — but this made me curious as to what Apple was doing for websites/Safari.

Turns out Apple may have done the API design for us already! In the WWDC 2023 session titled [“Meet Safari for spatial computing”](https://developer.apple.com/videos/play/wwdc2023/10279/) around 4:58 they address this gap by saying that Safari will provide the hover highlight effect 1) on elements with roles that suggest they’re interactive or — most importantly — elements with the CSS style `cursor: pointer`.

CleanShot 2024-01-10 at 13 21 06@2x

This is the API I think will best provide interoperability between all of the platforms I can think of, and makes the introduction of a new API to the core more palatable as its no longer specific to Apple platforms. Using `cursor: pointer` will be a clean match with the web and there are cursor APIs on Android and (I’m guessing) Windows/MacOS.

The only rub with this API design is that Apple’s API doesn’t just have one `UIHoverEffect` that is used, it has multiples and you’ve addressed that in your RFC but mine currently doesn’t. I need to spend the time to do more verification on the Vision Pro’s simulator but I’d guess that Safari assigns the interactive elements on a web page the `UIHoverAutomaticEffect` and that’s what I believe that React Native should map `cursor: pointer` too. Here’s a quick video of a prototype I had made taking this approach.

hover-demo

As for the other, more specific effects I think those should be platform specific and we should use “prefixed” values like the web does (at least they used to). The core can consider the `cursor` StyleSheet property to be a string value and if there are specific cursor styles that don’t fit into the keywords enumerated in the [web specification](https://drafts.csswg.org/css-ui/#cursor) the platform’s underlying view should listen for their own, platform prefixed values (and other platforms can ignore/no-op on values they don’t handle). As an example: on Apple platforms the two explicit, platform specific hover styles are `lift` and `highlight` so I think that if we want to expose them as an value for `cursor` they should be exposed as `-rn-apple-lift` and `-rn-apple-highlight`.

To begin with though (and to keep the scope of this RFC small) I’d suggest to at least start with just the `cursor: pointer` use-case and we can address the platform-specific values later as that as a separate RFC would probably include discussions on auxiliary gaps such as defining multiple/fallback values or having functionality equivalent to `@support` queries.

proposals/0000-hover-style-ipad-api.md
NickGerleman: Vincent left more comprehensive information about this below.

proposals/0000-hover-style-ipad-api.md
matthargett: I’m curious, what do you call these events for Windows Ink (Surface or Asus Pen)? Or for hand selection tracking in MRTK/HoloLens? I think there’s both prior art and probably a way to (finally) pull some of these offshoots forward in a unified way.

Saadnajmi:
@vincentriemer Thanks for the detailed response! I spoke with @okwasniewski offline and I think we both agree that `pointer:cursor` mapping to the automatic hover effect (or whatever Apple is doing for web) makes sense! A couple of extra thoughts:

1. React Native macOS (with help from Infinite Red and Meta's Messenger desktop team) previously implemented the `cursor` style prop, using as much of the W3C names as possible: https://github.com/microsoft/react-native-macos/pull/760. If the implementation is fine, maybe the approach we could take is to upstream part of that PR and combine it with @okwasniewski 's iOS/visionOS implementation. A caveat is that macOS supports [many more cursors](https://developer.apple.com/documentation/appkit/nscursor?language=objc) than iOS/iPad do, and I have no idea what Android does. Therefore, I think leaving the cursor prop as a string makes sense, with the caveat that maybe only the value `pointer` is supported in React Native iOS, while other platforms (like macOS/Windows/Web) may support a whole lot more? I don't know how far along your branch / draft is, I'm curious what you think of that?

2. For macOS, `cursor: pointer` adds the pointing hand cursor that I see used on web, but not actually in native Appkit experiences much. iPad also sparingly uses hover/pointer effects, I don't think they are intended to be on every button. For that reason, I think it might make sense to leave the prop off by default on all platforms, rather than my original thinking of having Pressable set `cursor: pointer` by default.

proposals/0000-hover-style-ipad-api.md
Saadnajmi: I have no idea what windows would call it 😅

okwasniewski:
Thanks, @vincentriemer @Saadnajmi @NickGerleman @matthargett for great feedback on this RFC. I've updated the proposed solution as @vincentriemer described in his comment.

proposals/0000-hover-style-ipad-api.md
NickGerleman: RN Windows solution for this is pretty crummy.

The event shape for custom `onMouseEnter`, `onMouseLeave` is effectively exact copy of [`PointerPoint`](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.input.pointerpoint?view=windows-app-sdk-1.4)

So, API is not coherent with other platforms. IIRC it may not even be typed.

NickGerleman: Pointer events unifies all of that, for I guess everything but Vision OS requiring declarative handling.

proposals/0000-hover-style-ipad-api.md
```diff
@@ -0,0 +1,101 @@
...
+
+    if (hoverEffect != nil) {
+        self.hoverStyle = [UIHoverStyle styleWithEffect:effect shape:shape];
+    }
+}
+```
+
+Here we initialize a new `UIShape` which reassembles the current’s view layer `cornerRadius`. Next we are creating a `UIHoverEffect` instance based on the value passed by user.
+
+The current API implementation might need some polish to correctly retrieve corner radius.
```
Saadnajmi, line 81: This would require some refactoring, but it would be nice if we could support the full custom border via separate-corner-radiuses-per-corner that RN supports (see `RCTBorderDrawing.h`). This is easier to do on Fabric than on Paper I think, but should be doable with both.

proposals/0000-hover-style-ipad-api.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Drawbacks
+
+- Apple only API
+  - Only Android equivalent I found is: https://developer.android.com/reference/android/view/MotionEvent which doesn’t provide any built-in effect but allows to create custom ones.
+- There might be different plan for the Pointer Events API.
```
Saadnajmi, line 87: I'd be curious to connect a mouse to an Android phone or tablet and see if the pointer ever changes around the system UI. I know you get a mouse cursor, I can test later. Regardless, this RFC should be apple-only to start IMO

vincentriemer: Android has a cursor that's more similar to web than Apple platforms via [this method on View](https://developer.android.com/reference/android/view/View.html#setPointerIcon(android.view.PointerIcon). I'd rather limit the RFC to apply to all platforms (at least theoretically) but limit the APIs surface area (like focusing on just `pointer: cursor` instead of all the different types).

okwasniewski:
@vincentriemer I agree that limiting this to `cursor: "pointer"` and `cursor: "auto"` and instead having this on all platforms is a better option. How can we move forward with this RFC? I could prepare a PR for iOS side of things. I could also try Android but I am not that familiar with it

Saadnajmi:
I've started working on an implementation of this RFC in both [React Native](https://github.com/facebook/react-native/pull/43078) (for just iOS) and [React Native macOS](https://github.com/microsoft/react-native-macos/pull/2080) (for visionOS, since macOS already had a cursor prop). Still rough but wanted to have my working branches up as we're already getting asks for this prop internally.

vincentriemer (approved):
lgtm

okwasniewski:
Hey! As this got implemented in https://github.com/facebook/react-native/pull/43078 I think this PR can be merged (?)

Saadnajmi:
While we're on the subject, are there opinions _against_ extending the "pointer-events" prop (not the new w3c api but the old one that controlled touch events) to also apply to cursors? That particular would help with some edge cases related to nested cursors, but I also know it's an old prop and may not be one we want to reuse. I think can be outside the scope of this rfc since the initial implementation already merged.

dlindenkreuz:
@Saadnajmi Great addition!

I haven't seen any discussion about [`UIPointerInteraction`](https://developer.apple.com/documentation/uikit/uipointerinteraction) yet, which has a slightly different appearance, is mostly used in iOS system apps, supports actual cursor styles (e.g. crosshair, i-beam) and is supported since iOS 13.4.

See this [WWDC 2020 video](https://developer.apple.com/videos/play/wwdc2020/10093/?time=792) (timestamped link) for more details.

Does it make sense to add support for `UIPointerInteraction` along the same lines?

Saadnajmi:
@dlindenkreuz I decided not to try `UIPointerInteraction` because:
1. The API was way more complex to implement
2. It _felt_ like Apple was replacing it with UIHoverStyle?

I didn't realize it allowed the other cursor styles though! I'm not sure it makes sense to have both `UIPointerStyle` and `UIHoverStyle` in the same view? Nevermind that more stuff is moving to SwiftUI..

dlindenkreuz:
> 1. The API was way more complex to implement

Agree.

There are four effect styles: [UIPointerEffect docs](https://developer.apple.com/documentation/uikit/uipointereffect)

https://developer.apple.com/documentation/uikit/pointer_interactions/integrating_pointer_interactions_into_your_ipad_app

I don't think a CSS rule `cursor: pointer` can represent the complexity of this system.

> 2. It felt like Apple was replacing it with UIHoverStyle?

Looks like `UIHoverStyle` is a convenient shorthand to `UIPointerInteraction` – I wouldn't be surprised if it uses a subset of `UIPointerInteraction` under the hood. Would be great to get >= iOS 13 support as well, maybe by replicating the functionality with `UIPointerInteraction`? 😬

PR #763, opened 2024-02-12
Support Image Symbol Names (e.g. SF Symbols)

getaaron:
This proposal suggests adding support for custom symbol images in React Native's `Image` component. Currently, React Native supports loading images from local files and network URLs. However, there's no direct support for using custom symbol images on iOS/macOS/tvOS. This feature would improve developers' ability to use React Native to create a native look and feel without relying on third-party libraries.

### DIFF
--- a/proposals/0000-support-image-symbol-names.md
+++ b/proposals/0000-support-image-symbol-names.md
@@ -1,0 +1,67 @@
+---
+title: Support Image Symbol Names
+author:
+- Aaron Brager
+date: 2024-02-11
+---
+
+# RFC0000: Support Image Symbol Names
+
+## Summary
+
+This proposal suggests adding support for custom symbol images in React Native's `Image` component. Currently, React Native supports loading images from local files and network URLs. However, there's no direct support for using custom symbol images on iOS/macOS/tvOS. This feature would improve developers' ability to use React Native to create a native look and feel without relying on third-party libraries.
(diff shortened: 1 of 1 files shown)


proposals/0000-support-image-symbol-names.md
```diff
@@ -0,0 +1,67 @@
...
+Android doesn't support an equivalent.
+
+## Drawbacks
+
+- Implementation cost: Adding support for custom symbol images may introduce additional complexity to the `Image` component implementation.
+- Platform specificity: This would only be available on Apple platforms which would exacerbate platform disparities in React Native apps.
+
+## Alternatives
+
+- Use `react-native-sfsymbols`
```
brentvatne, line 53: another good option: https://docs.expo.dev/versions/v51.0.0/sdk/symbols/

imo this is sufficient, in particular given https://github.com/react-native-community/discussions-and-proposals/pull/759 - which will see React Native move towards recommending using a React Native Framework like Expo and keeping a focused core. note that any platform-specific APIs in React Native core increase the surface area for out of tree platforms.

further, I'm not convinced that the `Image` component from react-native is actually the best tool to use for most apps. there is a greater discussion to be had about this, but in general most folks building larger apps will use expo-image or react-native-fast-image instead due to the better perceived "feel" and objective performance.

PR #787, opened 2024-05-17
RFC: Declare swift package manager dependencies in react-native library  podspec

mfazekas:
Proposes extension of cocapods utils in react-native to allow react-native libraries to declare SPM dependencies in their `podspec`

[View the rendered RFC](https://github.com/mfazekas/discussions-and-proposals/blob/spm-deps-in-podspec/proposals/0000-cocoapods-spm.md)


mfazekas:
0.75-rc.6 does have https://github.com/facebook/react-native/pull/44627 integrated now

PR #804, opened 2024-07-18
Add onPaste to TextInput

s77rt:
Proposal on adding `onPaste` prop to `TextInput` component to detect paste event. [View rendered proposal](https://github.com/s77rt/discussions-and-proposals/blob/TextInput-onPaste/proposals/0000-textinput-onpaste.md)

### DIFF
--- a/proposals/0000-textinput-onpaste.md
+++ b/proposals/0000-textinput-onpaste.md
@@ -1,0 +1,58 @@
+---
+title: Add onPaste to TextInput
+author:
+    - Abdelhafidh Belalia
+date: 2024-07-18
+---
+
+# RFC0000: Add onPaste to TextInput
+
+## Summary
+
+Add `onPaste` prop to `TextInput` component
+
+## Basic example
+
+```js
+<TextInput onPaste={() => console.log("User pressed Paste")} />
+```
+
+## Motivation
+
+Detect `Paste` event on TextInput. In our use case detecting paste event can be used to support image pasting.
+
+## Detailed design
+
+iOS:
+
+1. In `RCTUITextField`/`RCTUITextView` override the `paste:` method to call the delegate adapater `[didPaste]`
+2. Implement the `didPaste` method on the delegate adapater and call the delegate input `[textInputDidPaste]`
+3. Implement the `textInputDidPaste` method on the delegate input to disptach the `onPaste` event
+
+Android:
+
+1. In `ReactEditText` override the `onTextContextMenuItem` method to call the `PasteWatcher` insatnce
+2. In `ReactTextInputManager` create and set the `PasteWatcher` instance for the view (`ReactEditText`)
+3. Implement the `PasteWatcher` class to to disptach the `onPaste` event
+
+There is a PR for that change already https://github.com/facebook/react-native/pull/45425
+
+## Drawbacks
+
+-   onPaste callback timing is not the same in iOS vs Android. In iOS `onPaste` is called first then `onChange`. While in Android it's the other way around.
+
+## Alternatives
+
+None
+
+## Adoption strategy
+
+This is not a breaking change. Developers can adopt the feature by simply adding `onPaste` prop to their components
+
+## How we teach this
+
+Update the docs https://github.com/facebook/react-native-website/pull/4161
+
+## Unresolved questions
+
+-   How to resolve the timing inconsistency? (Not a blocker)
--- a/proposals/0000-textinput-onpaste.md
+++ b/proposals/0000-textinput-onpaste.md
@@ -14,7 +14,23 @@ Add `onPaste` prop to `TextInput` component
 ## Basic example
 
 ```js
-<TextInput onPaste={() => console.log("User pressed Paste")} />
+<TextInput
+	onPaste={(e) => {
+		const clipboardContent = e.nativeEvent.items[0];
+
+		if (clipboardContent.type === "text/plain") {
+			console.log(`Pasting text: ${clipboardContent.data}`);
+			return;
+		}
+
+		const file = {
+			uri: clipboardContent.data,
+			type: clipboardContent.type,
+		};
+		console.log("Uploading file");
+		uploadFile(file);
+	}}
+/>
 ```
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


NickGerleman:
On web, the analog of this event shape is `ClipboardEvent`, where we have cut/copy/paste events. https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent

Notably, the events expose `clipboardData`, with clipboard content cut or pasted. This may include more types than just text (like pasting an image, or a file), and may also be attached in more places (the API shares similarities to drag and drop).

With current proposal, the way we could try to read text changes (but only text) through `onChange`, but that may include full text including previous inputs. So, the component must now be controlled, and we must diff the string before/after to know what was pasted/where. Reading from clipboard global to try to get this may also be unreliably, since async.

Being able to read the underlying data which was pasted feels like an important part of this scenario (and browsers did ship it alongside their own paste events). So, partially modeling something like `clipboardData`, so the underlying text content can be exposed (and leaving us open to being able to listen for things like pasted files/images later).

proposals/0000-textinput-onpaste.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. In `ReactEditText` override the `onTextContextMenuItem` method to call the `PasteWatcher` insatnce
+2. In `ReactTextInputManager` create and set the `PasteWatcher` instance for the view (`ReactEditText`)
+3. Implement the `PasteWatcher` class to to disptach the `onPaste` event
+
+There is a PR for that change already https://github.com/facebook/react-native/pull/45425
+
+## Drawbacks
+
+-   onPaste callback timing is not the same in iOS vs Android. In iOS `onPaste` is called first then `onChange`. While in Android it's the other way around.
```
NickGerleman, line 42: Any `onChange`  triggered by paste would ideally come after paste event, per https://w3c.github.io/clipboard-apis/#otherevents

I think `onTextContextMenuItem()` might happen before change on Android?

s77rt: > I think `onTextContextMenuItem()` might happen before change on Android?

Indeed! I just found out that I was dispatching the event after calling `super.onTextContextMenuItem` (i.e. after pasting the text). Dispatching the event first fixes this inconsistency


https://github.com/user-attachments/assets/01f74705-90fe-40fb-8045-c1849aa83ea8

proposals/0000-textinput-onpaste.md
```diff
@@ -0,0 +1,58 @@
...
+
+This is not a breaking change. Developers can adopt the feature by simply adding `onPaste` prop to their components
+
+## How we teach this
+
+Update the docs https://github.com/facebook/react-native-website/pull/4161
+
+## Unresolved questions
+
+-   onPaste event does not expose any clipboard data. How can we add such data?
```
s77rt, line 58: @NickGerleman Can we handle this as a follow up?

NickGerleman: I don't think so. The event without data would require accessing clipboard afterward, and since that happens asynchronously, the data may have changed or be incorrect. A concrete example of this, is if you paste a password, a password manager might immediately clear it from the clipboard.

I.e. the current approach without the data is impossible to make reliable, and would encourage users to add code which falls into the anti-pattern.

I don't know how much time @cipolleschi has dedicated to this change, but did want to share the context I had on this, and some of the baselines of what these events should look like.

s77rt: The data can be as follow:
```js
{
    target: viewTag,
    items: [
        {
            type: "image/jpeg", // MIME type
            data: "data:image/jpeg;base64,..." // data: plaintext for text/plain and base64 encoded data for the rest
        }
    ],
}
```

items is an array that holds different items present in the clipboard, but for the current implementation only the first item is added.

Android implementation https://github.com/facebook/react-native/pull/45425/commits/af06d9ed1da0988654b8ba483024c203a5b7feb0


https://github.com/user-attachments/assets/808367fb-81f9-4f11-a4b5-d7c16a5207ea

s77rt: Regarding pasting files, using base64 string is not looking very performant (dealing with long strings can be slow). I'm looking into passing the URI instead, for Android we can directly use [`clipData.getItemAt(0).getUri()`](https://developer.android.com/reference/android/content/ClipData.Item#getUri())

(still investigating the same for iOS)

s77rt: On iOS the clipboard items do not expose any URI. When a file is pasted, we copy its data into a temporarily file and provide that file's URI instead. Proposal updated.

getsettalk:
this event is great `onPaste` this should be consider with high priorty

proposals/0000-textinput-onpaste.md
intergalacticspacehighway: @s77rt thanks for working on this. can we make the paste event resemble the `ClipboardEvent` API? Like this [example](https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event#examples)

So we can access clipboard data using `event.nativeEvent.clipboardData`. I like the temporary file approach over base64. Since we're returning the file URI, we can do below.

```
onPaste = (e) => {
    if (e.nativeEvent.clipboardData.types.indexOf('text/plain') > -1) {
        // User pasted a plain text
        const text = e.nativeEvent.clipboardData.getData('text/plain')
    }
    
    else if (e.nativeEvent.clipboardData.types.indexOf('text/uri-list') > -1) {
        // User pasted an image/file url
        const uri = e.nativeEvent.clipboardData.getData('text/uri-list')
        // in case of multiple URIs, they are separated by `\n` newline character.
    }
}
```

`clipboardData` is of type `DataTransfer` but for now we can support `types` and `getData`.

Also some more references - https://www.w3.org/TR/clipboard-apis/, [get data](https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-getdata)

blimpich:
@NickGerleman it looks like @s77rt's proposal [was updated](https://github.com/react-native-community/discussions-and-proposals/pull/804#discussion_r1706181000), do you think it needs further iterations before approval?

PR #813, opened 2024-08-12
Swift Turbo Modules

okwasniewski:
This PR describes API design that would allow Swift in Turbo Modules.

I've created this proposal together with @cipolleschi

During this effort I've been able to get Swift Turbo modules working (here is a native function call stopped on break point): 

image

# [A rendered version of the proposal can be read here](https://github.com/okwasniewski/discussions-and-proposals/blob/22edd2c7853aaf2ee654b57f4521e43d668103fb/proposals/0000-swift-turbo-modules.md)

### DIFF
--- a/proposals/0000-swift-turbo-modules.md
+++ b/proposals/0000-swift-turbo-modules.md
@@ -1,0 +1,185 @@
+---
+title: Swift Turbo Modules
+author:
+- Oskar Kwasniewski 
+- Riccardo Cipolleschi
+date: 12-08-2024 
+---
+
+# RFC0000: Swift Turbo Modules
+
+## Summary
+
+This RFC aims to allow developers to write Turbo Modules using Swift. This will allow the usage of more modern language making maintenance of native modules easier and more accessible. 
+
+## Motivation
+
+The primary motivations for introducing Swift Turbo Modules are:
+- Enhance developer experience for iOS developers working with React Native
+- Allow the use of more modern language
+- Make barrier to entry lower
+
+
+## Detailed design
+
+One of the problem why we can't adopt Swift in TurboModules  is the contamination of C++ ending up in user-space.
+
+### Current Situation
+The interfaces we generate for an Objective-C turbomodules have this shape:
+```objc
+@protocol NativeMyTurboModuleSpec <RCTBridgeModule, RCTTurboModule>
+@end
+```
+
+Where the `RCTTurboModule` protocol requires the conforming object to implement a method with this signature:
+```objc
+ - (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:
+       (const facebook::react::ObjCTurboModule::InitParams &)params;
+```
+
+As you can see, the signature of this method contains two types from C++:
+
+* `std::shared_ptr<facebook::react::TurboModule>`
+* `facebook::react::ObjCTurboModule::InitParams`
+
+### Solution
+
+The idea is to wrap the `getTurboModule` invocation in a `TurboModuleWrapper` object.
+
+The `TurboModuleWrapper` object is a base class that is supposed to be extended by a companion object for TurboModules. The base class has this interface:
+```objc
+@interface TurboModuleWrapper: NSObject
+- (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:
+    (const facebook::react::ObjCTurboModule::InitParams &)params;
+@end
+```
+And the base class implementation just fails, as we don't want for it to be used directly.
+
+Then, the `RCTTurboModule` interface can ask each TM to actually return an implementation of the `TurboModuleWrapper` rather then the actual `TurboModule`.
+
+The `TurboModuleWrapper` is a pure Objective-C class, so it will work seamlessly with ObjectiveC and Swift.
+
+So, now, the public interface of a TurboModule will only have pure objc entries and no C++ code.
+
+When it comes to the implementation, the user-defined TurboModule  won't have to deal with any C++ code:
+
+```objc
+@implementation MyTurboModule
+- (TurboModuleWrapper *)getWrapper
+{
+return [[MyTurboModuleWrapper alloc] init];
+}
+
+// ... rest of the TM methods ...
+
+@end
+```
+
+The `MyTurboModuleWrapper` implementation can be Codegenerated! 
+
+We don't even need to ask our users to write that code themselves, as we have all the informations we need in the Codegen already.
+
+
+The implementation will look like (note that `<UserDefinedName>` is something we get from Codegen):
+```objc
+// In the .h file
+
+@interface Native<UserDefinedName>Wrapper: TurboModuleWrapper
+@end
+
+// In the .mm file
+
+@implementation Native<UserDefinedName>Wrapper
+
+- (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:
+   (const facebook::react::ObjCTurboModule::InitParams &)params;
+{
+  return std::make_shared<facebook::react::Native<UserDefinedName>SpecJSI>(params);
+}
+
+@end
+```
+
+
+The user-defined TurboModule has already access to this file, and so the switch from `getTurboModule` to `getWrapper` doesn't require any additional includes.
+
+
+Finally, we will have to update the `RCTTurboModuleManager` to take this new object into consideration. So we have to modify the provide Turbomodule with the following code
+```objc
+  // Step 2e: Return an exact sub-class of ObjC TurboModule
+  std::shared_ptr<facebook::react::TurboModule> turboModule = nullptr;
+  
+  if ([module respondsToSelector:@selector(getTurboModule:)]) {
+    turboModule = [module getTurboModule:params];
+  } else if ([module respondsToSelector:@selector(getWrapper)]) {
+    auto wrapper = [module getWrapper];
+    turboModule = [wrapper getTurboModule:params];
+  }
+```
+
+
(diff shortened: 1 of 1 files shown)


proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+
+The primary motivations for introducing Swift Turbo Modules are:
+- Enhance developer experience for iOS developers working with React Native
+- Allow the use of more modern language
+- Make barrier to entry lower
+
+
+## Detailed design
+
+One of the problem why we can't adopt Swift in TurboModules  is the contamination of C++ ending up in user-space.
```
cipolleschi, line 25:
```suggestion
One of the reasons why we can't adopt Swift in TurboModules is the contamination of C++ types ending up in user-space.
```

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+## Summary
+
+This RFC aims to allow developers to write Turbo Modules using Swift. This will allow the usage of more modern language making maintenance of native modules easier and more accessible. 
+
+## Motivation
+
+The primary motivations for introducing Swift Turbo Modules are:
+- Enhance developer experience for iOS developers working with React Native
+- Allow the use of more modern language
+- Make barrier to entry lower
```
cipolleschi, line 20:
```suggestion
- Lower the entry barrier to write a native turbo module
```

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+One of the problem why we can't adopt Swift in TurboModules  is the contamination of C++ ending up in user-space.
+
+### Current Situation
+The interfaces we generate for an Objective-C turbomodules have this shape:
+```objc
+@protocol NativeMyTurboModuleSpec <RCTBridgeModule, RCTTurboModule>
+@end
+```
+
+Where the `RCTTurboModule` protocol requires the conforming object to implement a method with this signature:
```
cipolleschi, line 34:
```suggestion
The `RCTTurboModule` protocol requires the conforming object to implement a method with this signature:
```

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+```objc
+ - (std::shared_ptr<facebook::react::TurboModule>)getTurboModule:
+       (const facebook::react::ObjCTurboModule::InitParams &)params;
+```
+
+As you can see, the signature of this method contains two types from C++:
```
cipolleschi, line 40:
```suggestion
The signature of this method contains two types from C++:
```

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+
+* `std::shared_ptr<facebook::react::TurboModule>`
+* `facebook::react::ObjCTurboModule::InitParams`
+
+### Solution
+
+The idea is to wrap the `getTurboModule` invocation in a `TurboModuleWrapper` object.
```
cipolleschi, line 47:
```suggestion
The idea is to wrap the `getTurboModule` invocation in a `ModuleProvider` object.
```
or also

I think that Wrapper is not the right term for this pattern. we are calling a factory method to provide a new object, we are not wrapping a method call to adapt its interface or to decorate it with some other functionalities.

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+  }
+```
+
+
+Additionally we need to make sure that `React_Codegen` module is compatible with importing to Swift. I did a small test and adding few ifdefs to React_Codegen headers allows us to use Swift.
+
+```swift
+import protocol React_Codegen.NativeSwiftTestLibrarySpec
+import protocol React_Codegen.TurboModuleWrapper
+import class React_Codegen.NativeSwiftTestLibraryWrapper
```
cipolleschi, line 126:
```suggestion
import protocol ReactCodegen.NativeSwiftTestLibrarySpec
import protocol ReactCodegen.TurboModuleWrapper
import class ReactCodegen.NativeSwiftTestLibraryWrapper
```

Why not just importing `ReactCodegen`?

the `import protocol` and `import class` are kind of an advanced usage of the import statement in swift. Not many people use it or know what they are actually doing.

It could be that importing the whole `ReactCodegen` is not possible as it will make the build fail.

cipolleschi (commented):
Thanks @okwasniewski for putting this together and to start this conversation!

proposals/0000-swift-turbo-modules.md
okwasniewski: Exactly, at this point importing whole module is not possible but that's something we can adjust.

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+  if ([module respondsToSelector:@selector(getTurboModule:)]) {
+    turboModule = [module getTurboModule:params];
+  } else if ([module respondsToSelector:@selector(getWrapper)]) {
+    auto wrapper = [module getWrapper];
+    turboModule = [wrapper getTurboModule:params];
+  }
+```
+
+
+Additionally we need to make sure that `React_Codegen` module is compatible with importing to Swift. I did a small test and adding few ifdefs to React_Codegen headers allows us to use Swift.
```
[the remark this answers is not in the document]
okwasniewski: Fixed

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+	* The `TurboModule` itself can't adopt the `TurboModuleWrapper` protocol as, otherwise, the C++ signature will come back to the public API of the I don't think this will work. 
+	* We can't codegen the default implementation for the `getTurboModule` as we won't have the base class for the protocol.
+	* We could use a protocol and create a companion object in the codegen which extends the protocol and it is returned by the TurboModule. This solution works, but adds a bit of ceremonies to the base implementation above.
+
+- **Using a custom base class for each TurboModule**
+	The only pro of this approach is that we can remove a few lines from the definition of every TurboModule. 
+	The con of this approach are:
+	* The C++ code remains in the public API of the TurboModule
+	* We are creating a deeper inheritance chain which usually should be avoided.
+	* We are asking to all the users to inherit from a different base class. This is hard to make backward compatible.
```
TheRogue76, line 171: What about using the interop between Swift and C++? What is missing here right now when it comes to making modules? And how much work / breaking change would it need for it? I love this work, but going through ObjC is just going to mean every class still needs to deal with some level of ObjC nonsense, like the NSNumber stuff, and it is an additional layer between the code the library author writes and what is happening under the hood.

proposals/0000-swift-turbo-modules.md
```diff
@@ -0,0 +1,185 @@
...
+
+The primary motivations for introducing Swift Turbo Modules are:
+- Enhance developer experience for iOS developers working with React Native
+- Allow the use of more modern language
+- Lower the entry barrier to write a native turbo module
+
+
+## Detailed design
+
+One of the reasons why we can't adopt Swift in TurboModules is the contamination of C++ types ending up in user-space.
```
TheRogue76, line 25: While i can understand not wanting to expose internal C++ types to user land, reality is that RN is based on C++, and if something should be exposed and used by library maintainers, then having a clear API that defines public vs internal is all that is required, the language is inconsequential. I think as a library developer, regardless of what we do, we will eventually have to debug or deal with those internals, and this should be fine. Though maybe its an unpopular opinion

proposals/0000-swift-turbo-modules.md
okwasniewski: Currently, what's stopping Swift and C++ interop is the lack of virtual functions support. So this could be viable once Apple adds this feature. Another issue is that the C++ interop doesn't support having C++ in the header files, which is what React Native is doing. Codegen generates Objective-C interface anyways but we need a way of hiding C++ types from Swift and this is where the ModuleProvider comes in.

TheRogue76: Screenshot 2024-08-13 at 13 41 27
Based on https://developer.apple.com/videos/play/wwdc2024/10136/?time=1307 from WWDC24, the first one should not be an issue moving forward when Swift 6 drops.
As for the second one I am not sure what the usage is, can you explain that one a bit more?

cipolleschi: Hi there! 
There are several concerns here, to be honest.

What @okwasniewski explained is correct. I was in touch with Apple engineer and they mentioned that the inheritance model between Swift and C++ is too different to implement it properly, as of today. There are many cocnerns related to what a virtual class implies in C++ and a protocol in Swift. TL;DR: it will not happen that Swift class can inherit from C++.

On another side, Swift/Cxx interop is too young and unstable to be used concretely, imho.
I'm very scared that if we start introducing it, people that have custom project might incur in weird compilation issues and that maintenance cost will fall on our shoulder.
Instead, objc/swift interop is well documented and supported and, overall, more stable.

I agree that is perhaps non ideal in terms of ergonomics, but we can try to mitigate the problem Codegenerating as much as we can.

proposals/0000-swift-turbo-modules.md
cipolleschi: I completely agree with you. But we would like to provide a cleaner and simple experience to library maintainers. Not every library needs to know how react native works under the hoods, no?

We also have an effort internally to try and define the public boundary better, with a clear distinction between what's public and what's private. The project is almost 10 years old in OSS, so work is not trivial and requires a bit of time, but we hope to get there soon!

PR #826, opened 2024-10-15
Add regex to TextInput

s77rt:
Proposal on adding `regex` prop to `TextInput` component to restrict user input. [View rendered proposal](https://github.com/s77rt/discussions-and-proposals/blob/TextInout-regex/proposals/0000-textinput-regex.md)


coolsoftwaretyler:
I think this would be very useful! Seems a lot like the pattern attribute in HTML.

PR #947, opened 2025-10-10
RFC: ArrayBuffer support in TurboModules

paradowstack:
Proposal: Adding first-class `ArrayBuffer` support to Codegen and TurboModules to enable zero-copy binary data exchange between JavaScript and Native modules.

# **[View the rendered RFC](https://github.com/paradowstack/discussions-and-proposals/blob/feat/arraybuffer-in-turbomodules/proposals/0000-arraybuffer-support-in-turbomodules.md)**

### DIFF
--- a/proposals/0000-arraybuffer-support-in-turbomodules.md
+++ b/proposals/0000-arraybuffer-support-in-turbomodules.md
@@ -1,0 +1,381 @@
+---
+title: ArrayBuffer support in TurboModules
+author:
+  - Kamil Paradowski <kamil.paradowski@callstack.com>
+date: 10.10.2025
+---
+
+# RFC0000: ArrayBuffer support in TurboModules
+
+## Summary
+
+This RFC outlines adding first-class `ArrayBuffer` support to TurboModules to enable zero-copy binary data exchange between JavaScript and native modules.
+
+Codegen and TurboModules currently lack efficient binary data handling, forcing developers to rely on workarounds, such as using the `CodegenTypes.UnsafeObject` type or inefficient conversion to `base64`, when working with binary-heavy use cases. This feature addresses performance bottlenecks in data-intensive applications. By providing first-class `ArrayBuffer` support, developers will be able to seamlessly pass large binary payloads between JavaScript and native code without the overhead of data serialization and copying. This enhancement will unlock new possibilities for creating responsive, high-performance React Native applications across various domains including media processing, real-time communication, and AI/ML applications that require efficient handling of tensors and model data.
+
+## Basic example
+
+Developers should be able to define a TypeScript spec as follows:
+
+```ts
+export interface Spec extends TurboModule {
+  getBuffer(): ArrayBuffer;
+  processBuffer(buffer: ArrayBuffer): void;
+}
+```
+
+## Motivation
+
+TurboModules currently lack a first-class way to represent `ArrayBuffer` end-to-end in Codegen, which forces developers to rely on copies, ad-hoc platform bridges, global helpers, or external libraries. This hurts performance for binary-heavy use cases such as media data or ML tensors, and it increases implementation complexity. The expected outcome is a cross-platform contract that lets JS and native pass binary data with minimal copying. Codegen should be able to generate working code for the `ArrayBuffer` type on every platform.
+
+For example, several important use cases are currently difficult to implement efficiently while working with TurboModules:
+
+- **Real-time media streaming**: A native video decoder could stream frames directly to a JavaScript-based player component. Without zero-copy `ArrayBuffer`s, each frame would need to be copied, leading to significant performance overhead and potential frame drops.
+- **Machine Learning**: On-device ML models often require passing large tensors between native inference engines and JS. Copying this data can be a major bottleneck, especially for real-time applications like video analysis.
+- **High-performance networking**: Applications that handle large binary payloads over WebSockets or other protocols (e.g., financial data streams, real-time gaming) may be forced into inefficient data conversion, which adds CPU and memory pressure.
+
+By providing a first-class `ArrayBuffer` type support to TurboModules, this RFC will unblock these and other performance-sensitive areas, making it possible to develop faster more efficient applications for React Native.
+
+## Detailed design
+
+This section contains a description of the design and related topics. It's split by topics and affected packages/areas. In the expandable sections, extensive code snippets can be found. In the final implementation, each area should come with appropriate unit tests.
+
+### Memory ownership
+
+When passing an `ArrayBuffer` to native code, it should always be treated as "borrowed" or "non-owning": JS owns the ArrayBuffer's memory and the JS GC is responsible for freeing it. Native code should access the passed memory only for the duration of the synchronous call.
+
+The same rules apply when passing buffers from native to JS: native code remains the owner of the allocated memory and may expose zero-copy buffers to JS.
+
+### Threading
+
+`ArrayBuffer` implementations are not thread-safe; if multiple threads simultaneously read from or write to an `ArrayBuffer`, race conditions can occur. To prevent this, developers must ensure that an `ArrayBuffer` is not accessed concurrently from different threads, for example by making sure that JavaScript thread does not modify the `ArrayBuffer` while native code is working on it.
+
+### Types
+
+To address this gap in Codegen, we first need to specify the corresponding native type for every language used by TurboModules. The type should have the following traits:
+
+- Provide access to the raw memory of the buffer
+- Provide the length of the buffer
+- Provide a way to mutate/modify the buffer contents
+- Provide a zero-copy construction path without extra allocations
+- Not take ownership of the underlying memory block
+
+Results can be found in the table below.
+
+| Language    | Type                  |
+| ----------- | --------------------- |
+| JavaScript  | `ArrayBuffer`         |
+| C++         | `jsi::ArrayBuffer`    |
+| Java        | `java.nio.ByteBuffer` |
+| Objective-C | `NSMutableData`       |
+
+We need types that provide zero-copy semantics and clear ownership of the underlying data. Fortunately, we can reuse existing native types instead of implementing custom ones. Below is a breakdown by platform.
+
+#### C++
+
+The C++ case requires the fewest changes and least discussion when choosing the native type. JSI already has first-class support for ArrayBuffers. The JSI `jsi::Object` can hold an ArrayBuffer instance. `jsi::Object` provides `ArrayBuffer getArrayBuffer(Runtime& runtime)` and `bool isArrayBuffer(Runtime& runtime)`, which make it easy to convert objects to buffers. `jsi::ArrayBuffer` provides access to internal data and its size. Moreover, this class inherits from `jsi::Object`, can be implicitly converted to its parent class, and is intended to be constructed in a non-copying way (using move semantics or by passing a shared pointer to `jsi::MutableBuffer`).
+
+However, to maintain consistency with other types, we should add a helper `ArrayBuffer asArrayBuffer(Runtime& runtime)` to the `jsi::Object` class.
+
+#### Java
+
+Java provides a class that matches our needs: `java.nio.ByteBuffer`. It offers the required functionality and can be created via `ByteBuffer.allocateDirect()` or `NewDirectByteBuffer()` to wrap existing memory without extra allocations — unlike a `byte[]` alternative. It also provides access to the raw data pointer, individual bytes, and its length.
+
+#### Objective-C
+
+For Objective-C we chose `NSMutableData` for the same reasons described above. It allows creating an object that wraps existing memory and clearly defines ownership of the memory block.
+
+### Zero-copy conversion
+
+Below you can find required changes and example zero-copy conversion implementations for all platforms.
+
+#### C++
+
+No changes are needed in JSI. Its in-place support for `ArrayBuffer` already provides the necessary in-place conversion functionality.
+
+#### Java
+
+Java class `java.nio.ByteBuffer` can be constructed using JNI function `jobject NewDirectByteBuffer(JNIEnv* env, void* address, jlong capacity)`. This function returns a direct instance of direct `java.nio.ByteBuffer` referring to the block of memory starting at the memory address `address` and extending `capacity` bytes. What is important, direct buffers don't deallocate the memory on destruction, what is desired in our case since JS is responsible for that.
+
+<details>
+<summary>Example conversion implementation</summary>
+
+`packages/react-native/ReactCommon/react/nativemodule/core/platform/android/ReactCommon/JavaTurboModule.cpp`:
+
+```cpp
+  // jsi::ArrayBuffer -> java.nio.ByteBuffer
+  if (type == "Ljava/nio/ByteBuffer;") {
+    if (!(arg->isObject() && arg->getObject(rt).isArrayBuffer(rt))) {
+      throw JavaTurboModuleArgumentConversionException(
+          "ArrayBuffer", argIndex, methodName, arg, &rt);
+    }
+    auto arrayBuffer = arg->asObject(rt).asArrayBuffer(rt);
+    auto len = arrayBuffer.size(rt);
+    auto data = arrayBuffer.data(rt);
+    auto directBuffer = env->NewDirectByteBuffer(
+        static_cast<void*>(data), static_cast<jlong>(len));
+    jarg->l = makeGlobalIfNecessary(directBuffer);
+    continue;
+  }
+  ...
+  // java.nio.ByteBuffer -> jsi::ArrayBuffer
+  case ArrayBufferKind: {
+    auto returnObject =
+        (jobject)env->CallObjectMethodA(instance, methodID, jargs.data());
+    checkJNIErrorForMethodCall();
+
+    TMPL::syncMethodCallExecutionEnd(moduleName, methodName);
+    TMPL::syncMethodCallReturnConversionStart(moduleName, methodName);
+
+    auto returnValue = jsi::Value::null();
+    if (returnObject != nullptr) {
+      auto jResult = jni::adopt_local(returnObject);
+      auto byteBuffer = jResult.get();
+
+      struct ByteArrayMutableBuffer : jsi::MutableBuffer {
+        public:
+           ByteArrayMutableBuffer(uint8_t* data, size_t size)
+               : _data{data}, _size{size} {}
+
+           uint8_t* data() override {
+             return _data;
+           }
+           size_t size() const override {
+             return _size;
+           }
+
+        private:
+           uint8_t* _data{};
+           size_t _size{};
+      };
+
+      auto size = env->GetDirectBufferCapacity(byteBuffer);
+      auto data = (uint8_t*)env->GetDirectBufferAddress(byteBuffer);
+      auto mutableBuffer =
+          std::make_shared<ByteArrayMutableBuffer>(data, size);
+      auto arrayBuffer = jsi::ArrayBuffer{runtime, mutableBuffer};
+      auto obj = jsi::Value{runtime, arrayBuffer};
+      returnValue = std::move(obj);
+    }
+
+    TMPL::syncMethodCallReturnConversionEnd(moduleName, methodName);
+    TMPL::syncMethodCallEnd(moduleName, methodName);
+    return returnValue;
+  }
+```
+
+> [!NOTE]
+> In the final implementation temporary structure `ByteArrayMutableBuffer` can be moved outside of the function.
+
+</details>
+
+#### Objective-C
+
+In Objective-C, we need conversions between `facebook::jsi::ArrayBuffer` and `NSMutableData`. Creating an instance of `NSMutableData` wrapping the memory of the JS `ArrayBuffer` without taking ownership of it can be achieved by calling the `dataWithBytesNoCopy` function with `freeWhenDone` set to `NO`.
+
+<details>
+<summary>Example conversion implementation</summary>
+
+`packages/react-native/ReactCommon/react/nativemodule/core/platform/ios/ReactCommon/RCTTurboModule.mm`:
+
+```objc
+// NSData -> jsi::ArrayBuffer
+static jsi::ArrayBuffer convertNSDataToJSIArrayBuffer(jsi::Runtime &runtime, NSData *value)
+{
+	auto length = [value length];
+	auto buffer = (uint8_t*)[value bytes];
+
+	class NSDataMutableBuffer : public facebook::jsi::MutableBuffer {
+	 public:
+		NSDataMutableBuffer(uint8_t* data, size_t size)
+			: _data{data}, _size{size} {}
+
+		uint8_t* data() override {
+			return _data;
+		}
+		size_t size() const override {
+			return _size;
+		}
+
+	 private:
+		uint8_t* _data{};
+		size_t _size{};
+	};
+
+	auto mutableBuffer = std::make_shared<NSDataMutableBuffer>(buffer, length);
+	auto arrayBuffer = jsi::ArrayBuffer(runtime, mutableBuffer);
+	return arrayBuffer;
+}
+...
+// jsi::ArrayBuffer -> NSData
+static NSMutableData *convertJSIArrayBufferToNSData(
+		jsi::Runtime &runtime,
+		const jsi::ArrayBuffer &value,
+		const std::shared_ptr<CallInvoker> &jsInvoker,
+		BOOL useNSNull)
+{
+	auto length = value.size(runtime);
+	auto buffer = value.data(runtime);
+
+	NSMutableData* data = [NSMutableData dataWithBytesNoCopy:buffer length:length freeWhenDone:NO];
+
+	return data;
+}
+```
+
+> [!NOTE]
+> In the final implementation temporary structure `NSDataMutableBuffer` can be moved outside of the function.
+
+</details>
+
+### Usage
+
+The end goal of this RFC is to enable developers to use the `ArrayBuffer` type in spec files, use Codegen to generate native interfaces and glue code, and implement TurboModule functions that accept or return `ArrayBuffer`.
+
+#### TypeScript
+
+```typescript
+export interface Spec extends TurboModule {
+  getBuffer(): ArrayBuffer;
+  processBuffer(buffer: ArrayBuffer): void;
+}
+```
+
+#### C++
+
+```cpp
+class JSI_EXPORT NativeBufferCxxSpecJSI : public TurboModule {
+protected:
+  NativeBufferCxxSpecJSI(std::shared_ptr<CallInvoker> jsInvoker);
+
+public:
+  virtual jsi::ArrayBuffer getBuffer(jsi::Runtime &rt) = 0;
+  virtual jsi::Value generateAsyncBuffer(jsi::Runtime &rt) = 0;
+  virtual jsi::String processBuffer(jsi::Runtime &rt, jsi::ArrayBuffer buffer) = 0;
+};
+```
+
+#### Java
+
+```java
+public abstract class NativeBufferSpec extends ReactContextBaseJavaModule implements TurboModule {
+  ...
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  @DoNotStrip
+  public abstract java.nio.ByteBuffer getBuffer();
+
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  @DoNotStrip
+  public abstract String processBuffer(java.nio.ByteBuffer buffer);
+}
+
+```
+
+#### Objective-C
+
+```objc
+@protocol NativeBufferSpec <RCTBridgeModule, RCTTurboModule>
+
+- (NSMutableData *)getBuffer;
+- (NSString *)processBuffer:(NSMutableData *)buffer;
+
+@end
+```
+
+### Scope of `ArrayBuffer` support for Codegen
+
+So far this RFC mentioned support for `ArrayBuffer` only for a specific range of use cases. As can be noticed in the section above, these are:
+
+- Function output argument (`getBuffer`)
+- Function input argument (`processBuffer`)
+
+However, Codegen provides more functionality and broader support for types that can be shared across platforms. Below you can find the list (probably not full) of them:
+
+- Promises
+
+```ts
+export interface Spec extends TurboModule {
+  getBuffer(): ArrayBuffer;
+  getAsyncBuffer(): Promise<ArrayBuffer>;
+}
+```
+
+- Structs
+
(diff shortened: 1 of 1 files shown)


kraenhansen:
Firstly - congrats on a very thorough and well written RFC!

> do we need to implement any additional synchronization mechanism to make this change thread-safe?

It's my understanding that the JS engine and JSI itself isn't thread-safe either.
In which case, I'd consider the limitation of thread-safely for `ArrayBuffer` an extension of that.

> Should we introduce two kinds of buffers, mutable and read-only

I think that could be a nice addition indeed. I'd actually imagine a read-only variant would be used in most cases 🤔

> Should this RFC focus on introducing only the basic and most valuable synchronous support for ArrayBuffer

Personally - I'd apply the 80-20 rule here and go for the least amount of work bringing most of the value and stick with the basic sync support.

---

Did you consider "views"? ([`DataView`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) and [typed arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)) and how those would interact with this feature? Could these be passed between JS and Native, if not - what's the failure case like? And should typed arrays be supported in codegen?

gmemmy:
i really like this. for real-time media or GPU pipelines, say camera ---> ML ---> WebRTC, efficient ArrayBuffer bridging would make a world of difference. it'd enable moving small binary payloads(LUTs, masks, uniform buffers) without having to serialize or clone data just to cross the bridge.

a few things worth clarifying though:

- thread safety: how do concurrent module calls handle shared buffers? media pipelines often run multiple workers in parallel, so locking semantics matter.

- read-only vs writable: some data(e.g frame masks or GPU uploads) should likely be immutable once passed to JS; being explicit here could save a lot of edge-case debugging imo.

paradowstack:
> Firstly - congrats on a very thorough and well written RFC!

Thanks ❤️ 
 
> > do we need to implement any additional synchronization mechanism to make this change thread-safe?
> 
> It's my understanding that the JS engine and JSI itself isn't thread-safe either. In which case, I'd consider the limitation of thread-safely for `ArrayBuffer` an extension of that.

Agree with that.
 
> > Should we introduce two kinds of buffers, mutable and read-only
> 
> I think that could be a nice addition indeed. I'd actually imagine a read-only variant would be used in most cases 🤔

I agree with that, but after deeper investigation I couldn't find an easy and clean way to achieve that. One way would be to create a read-only view over the buffer when processing it, but that's on the developers. Also there is an active [TC39 proposal for an Immutable ArrayBuffer](https://tc39.es/proposal-immutable-arraybuffer/) which would provide a standardized, runtime-enforced way to prevent modifications to the buffer contents.
 
> > Should this RFC focus on introducing only the basic and most valuable synchronous support for ArrayBuffer
> 
> Personally - I'd apply the 80-20 rule here and go for the least amount of work bringing most of the value and stick with the basic sync support.

👍 
 
> Did you consider "views"? ([`DataView`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) and [typed arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)) and how those would interact with this feature? Could these be passed between JS and Native, if not - what's the failure case like? And should typed arrays be supported in codegen?

My idea is to have this solution type-agnostic as the underlying native classes, such as `NSMutableData`, `java.nio.ByteBuffer` and `jsi::ArrayBuffer` are an opaque containers for raw, uninterpreted bytes. If `DataView` or `TypedArray` is passed instead of `ArrayBuffer`, the developer should receive an appropriate warning.

tom-sherman:
The semantics around memory ownership seem to deviate from the spec of ArrayBuffer in regards to transferring/detaching. I'm not sure of what the material consequences are, especially with existing code that handles ArrayBuffers, but it seems like this could break developer expectations in many ways.

I'd expect the buffer to be moved, not borrowed, when passing between JS and native (in both directions).

> `ArrayBuffer` implementations are not thread-safe; if multiple threads simultaneously read from or write to an `ArrayBuffer`, race conditions can occur. To prevent this, developers must ensure that an `ArrayBuffer` is not accessed concurrently from different threads

This problem goes away if ownership is moved to the receiving thread. You shouldn't be able to even read an ArrayBuffer from multiple threads.

> 1.  Thread-safety of the `ArrayBuffer` - do we need to implement any additional synchronization mechanism to make this change thread-safe?

So in summary and to answer this unresolved question, I would say absolutely yes. And to do so by using moves and not borrows.

paradowstack:
Thanks @tom-sherman  for your input!

Regarding this:

> > `ArrayBuffer` implementations are not thread-safe; if multiple threads simultaneously read from or write to an `ArrayBuffer`, race conditions can occur. To prevent this, developers must ensure that an `ArrayBuffer` is not accessed concurrently from different threads
> 
> This problem goes away if ownership is moved to the receiving thread. You shouldn't be able to even read an ArrayBuffer from multiple threads.
> 
> > 1. Thread-safety of the `ArrayBuffer` - do we need to implement any additional synchronization mechanism to make this change thread-safe?
> 
> So in summary and to answer this unresolved question, I would say absolutely yes. And to do so by using moves and not borrows.

I agree that moving (transferring ownership) an ArrayBuffer coul be fundamentally safer and cleaner than borrowing it. However, the primary technical challenge remains: the current JSI and Hermes Runtime implementations do not expose a dedicated API for "detaching" an ArrayBuffer from the JavaScript side. Without true detachment, the only immediate way to transfer ownership is by "moving" the underlying buffer to the native thread and extending its lifetime accordingly. This addresses the memory management aspect but has a critical flaw:
- **JS Validity**: The ArrayBuffer remains valid on the JS side. Its properties, such as `byteLength`, are not cleared.
- **Thread Safety Risk**: The buffer can still be read or written to simultaneously from the JS thread while it is being used natively. This creates an easy opportunity for developers to violate thread-safety rules, even if documentation warns against post-transfer access.

I currently do not see a clean, safe path to fully implement buffer transfers that invalidate the JS reference. Achieving this requires dedicated changes to both the JSI specification and the underlying Hermes engine to introduce a proper detachment mechanism. Since I don't have a deep expertise in this topic, output from more experienced developers is really welcome and highly appreciated.

gmaclennan:
I don't have any expertise as to how to solve the invalidation of JS references in Hermes and JSI, but I wanted to add another voice highlighting the importance of ownership transfer. As far as I am aware, JS does not have other instances where a developer needs to think about thread-safety - it is always thread safe by default. When working with threads (e.g. a `Worker` in node or a `WebWorker` in the browser), references are either copied or transferred (e.g. zero-copy, but the reference is removed in the source thread). A JS developer needing to be aware that they cannot modify an ArrayBuffer while it is being written in the native thread is a big ask, especially as we are likely talking about developers who are consuming native modules which use this code, and may not be familiar with thread-safety as a concept at all. This might be a really hard problem to solve though as @paradowstack says!

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+Codegen and TurboModules currently lack efficient binary data handling, forcing developers to rely on workarounds, such as using the `CodegenTypes.UnsafeObject` type or inefficient conversion to `base64`, when working with binary-heavy use cases. This feature addresses performance bottlenecks in data-intensive applications. By providing first-class `ArrayBuffer` support, developers will be able to seamlessly pass large binary payloads between JavaScript and native code without the overhead of data serialization and copying. This enhancement will unlock new possibilities for creating responsive, high-performance React Native applications across various domains including media processing, real-time communication, and AI/ML applications that require efficient handling of tensors and model data.
+
+## Basic example
+
+Developers should be able to define a TypeScript spec as follows:
+
+```ts
+export interface Spec extends TurboModule {
+  getBuffer(): ArrayBuffer;
+  processBuffer(buffer: ArrayBuffer): void;
```
grabbou, line 23: Should we extend the RFC with some notes on asynchronous functions? I know there was discussion beforehand whether to implement asynchronous code in the first PR, however, I would treat that separate from the RFC itself, which could cover broader use case.

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+}
+```
+
+## Motivation
+
+TurboModules currently lack a first-class way to represent `ArrayBuffer` end-to-end in Codegen, which forces developers to rely on copies, ad-hoc platform bridges, global helpers, or external libraries. This hurts performance for binary-heavy use cases such as media data or ML tensors, and it increases implementation complexity. The expected outcome is a cross-platform contract that lets JS and native pass binary data with minimal copying. Codegen should be able to generate working code for the `ArrayBuffer` type on every platform.
```
grabbou: I would say current workaround is typically working directly with JSI, which has its own advantages and disadvantages. Great example here is prior art by Marc, who did a lot of this manually before migrating over to Nitro Modules. I also think a lot of Software Mansion libraries go with C++ and work directly with JSI for that reason as well.

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+
+For example, several important use cases are currently difficult to implement efficiently while working with TurboModules:
+
+- **Real-time media streaming**: A native video decoder could stream frames directly to a JavaScript-based player component. Without zero-copy `ArrayBuffer`s, each frame would need to be copied, leading to significant performance overhead and potential frame drops.
```
grabbou, line 37: I think great example here would be to use Blob Manager as an example (and likely first candidate to migrate over, once this lands). 

Broadly speaking working with binary data in general.

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+- **Machine Learning**: On-device ML models often require passing large tensors between native inference engines and JS. Copying this data can be a major bottleneck, especially for real-time applications like video analysis.
+- **High-performance networking**: Applications that handle large binary payloads over WebSockets or other protocols (e.g., financial data streams, real-time gaming) may be forced into inefficient data conversion, which adds CPU and memory pressure.
+
+By providing a first-class `ArrayBuffer` type support to TurboModules, this RFC will unblock these and other performance-sensitive areas, making it possible to develop faster more efficient applications for React Native.
```
grabbou: Another important indication here is that first-class support for ArrayBuffers will enable support of web spec compatible primitives (such as Blob or File). It is also great way to enable interoperability across different module frameworks.

For example, Expo currently has its own Blob implementation (based on ArrayBuffers, which they support). If React Native has same level of support, libraries relying on each will be able to work interchangeably, which is great.

grabbou (commented):
Thanks for this RFC. Excited to get this going. Leaving some comments for consideration when it comes to the RFC document itself.

grabbou:
> The semantics around memory ownership seem to deviate from the spec of ArrayBuffer in regards to transferring/detaching. I'm not sure of what the material consequences are, especially with existing code that handles ArrayBuffers, but it seems like this could break developer expectations in many ways.

That said, it's worth noting that (unless I'm mistaken) this is already how JSI, Expo Modules, and Nitro Modules handle ArrayBuffers today.

On one hand, aligning with existing community behavior might make sense for practical and compatibility reasons. On the other hand, once this behavior becomes part of the core, its reach and visibility will likely expand far beyond those ecosystems, making the current de facto behavior less relevant over time.

Leaving this as an open question and summoning a few folks from the community for feedback!

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+#### Objective-C
+
+No obstacles were found for this platform.
+
+#### Conclusion
+
+To conclude, supporting only simple function input/output parameters is straightforward, but extending that support to cover all Codegen functionalities across every platform is significantly more complex. However, integrating Promises and asynchronous operations may be crucial for supporting binary-heavy use cases. This leads to the question:
+
+> [!IMPORTANT]
+> Should this RFC focus on introducing only the basic and most valuable synchronous support for ArrayBuffer, or should it aim for full coverage of all possible Codegen use cases, including asynchronous operations, despite the higher complexity and the impact on more files (especially on Android)?
```
grabbou: In my opinion, the RFC should aim to cover all use cases and describe the complete implementation plan. We can then approach it incrementally, breaking the work into smaller, manageable PRs. It might also make sense to update the Adoption Strategy section with a detailed roll-out plan that outlines the milestones we’ll follow once this RFC is approved.

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+
+Below you can find required changes and example zero-copy conversion implementations for all platforms.
+
+#### C++
+
+No changes are needed in JSI. Its in-place support for `ArrayBuffer` already provides the necessary in-place conversion functionality.
+
+#### Java
+
+Java class `java.nio.ByteBuffer` can be constructed using JNI function `jobject NewDirectByteBuffer(JNIEnv* env, void* address, jlong capacity)`. This function returns a direct instance of direct `java.nio.ByteBuffer` referring to the block of memory starting at the memory address `address` and extending `capacity` bytes. What is important, direct buffers don't deallocate the memory on destruction, what is desired in our case since JS is responsible for that.
```
javache, line 138: Would recommend we use the existing fbjni abstraction here: https://github.com/facebookincubator/fbjni/blob/main/cxx/fbjni/ByteBuffer.cpp#L84

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+
+## Detailed design
+
+This section contains a description of the design and related topics. It's split by topics and affected packages/areas. In the expandable sections, extensive code snippets can be found. In the final implementation, each area should come with appropriate unit tests.
+
+### Memory ownership
+
+When passing an `ArrayBuffer` to native code, it should always be treated as "borrowed" or "non-owning": JS owns the ArrayBuffer's memory and the JS GC is responsible for freeing it. Native code should access the passed memory only for the duration of the synchronous call.
+
+The same rules apply when passing buffers from native to JS: native code remains the owner of the allocated memory and may expose zero-copy buffers to JS.
```
javache: Why is this so? `createArrayBuffer` uses a `shared_ptr` so it should be feasible to implemented shared ownership semantics here.

javache:
Thanks for putting this together.

Generally, I'm very aligned supporting a type-safe abstraction over the existing ArrayBuffer support in JSI, and there seems to be plenty of use-cases where this would become a good way forward to unlock cheaper data sharing between JS and native.

I agree with the concerns around thread-safety expressed in this thread here. Is there any prior art we can reference? Is the operation model similar to [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer)? Should we consider [Atomics](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics) as a complementary but necessary capability here?

Making this fully support async JS to native invocation calls will likely increase complexity, as it would require us to keep the JS object alive for the duration of the native memory reference, but not impossible. Alternatively, we'd need to make it really obvious that ArrayBuffer args can only be used in sync calls through codegen.

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+
+By providing a first-class `ArrayBuffer` type support to TurboModules, this RFC will unblock these and other performance-sensitive areas, making it possible to develop faster more efficient applications for React Native.
+
+## Detailed design
+
+This section contains a description of the design and related topics. It's split by topics and affected packages/areas. In the expandable sections, extensive code snippets can be found. In the final implementation, each area should come with appropriate unit tests.
+
+### Memory ownership
+
+When passing an `ArrayBuffer` to native code, it should always be treated as "borrowed" or "non-owning": JS owns the ArrayBuffer's memory and the JS GC is responsible for freeing it. Native code should access the passed memory only for the duration of the synchronous call.
```
mrousavy: Seems like this terminology is from Nitro 😄

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,381 @@
...
+
+#### C++
+
+It looks like the current implementation of `AsyncEventEmitter` requires an argument type to be copyable. That is not the case for `jsi::ArrayBuffer`. This will probably require deeper investigation and perhaps more complex changes to the bridging part of the `ReactCommon` library.
+
+#### Java
+
+While in C++ and Objective-C data can be easily shared between JS and Native, Java stores the data as a `folly::dynamic` map on the Native side. The `folly` library has support for data buffers (class `IOBuf`). This means that JS buffers can be stored on the Native side, but its implementation will be more challenging. Moreover, classes responsible for storing variables, such as `NativeMap` or `NatviveArray`, have a rich inheritance tree and are widely used across the JNI files. Adding storage for buffers to them will require changes to a large number of `ReactAndroid` JNI and Java/Kotlin files. These changes are required to add support for e.g. Promises or Structs.
```
mrousavy:
```suggestion
While in C++ and Objective-C data can be easily shared between JS and Native, Java stores the data as a `folly::dynamic` map on the Native side. The `folly` library has support for data buffers (class `IOBuf`). This means that JS buffers can be stored on the Native side, but its implementation will be more challenging. Moreover, classes responsible for storing variables, such as `NativeMap` or `NativeArray`, have a rich inheritance tree and are widely used across the JNI files. Adding storage for buffers to them will require changes to a large number of `ReactAndroid` JNI and Java/Kotlin files. These changes are required to add support for e.g. Promises or Structs.
```

mrousavy: > While in C++ and Objective-C data can be easily shared between JS and Native, Java stores the data as a `folly::dynamic` map on the Native side.

Yea I found the same issue a while ago - folly is quite cool but wrapping everything in dynamic probably has to go at some point in the future.

mrousavy (approved):
I really like this RFC - adding `ArrayBuffer` to react-native core seems like a no-brainer.

I can see that some of this was inspired by [Nitro](https://nitro.margelo.com), where we already have `nitro::ArrayBuffer` (see [the `ArrayBuffer` docs](https://nitro.margelo.com/docs/types/array-buffers)).

I remember implementing `ArrayBuffer`s wasn't super easy, as [a lot of work went into the `nitro::ArrayBuffer` primitives](https://github.com/mrousavy/nitro/commits/main/packages/react-native-nitro-modules/cpp/core/ArrayBuffer.hpp) - I think I learned a lot about how this works - I can help with the implementation if any questions arise.

Nitro has **owning** and **non-owning** ArrayBuffers, where **owning** is a native one (we can unwrap that later again), and **non-owning** is one that was created in JS, or somewhere else.

**PROBLEM**: If an **owning** ArrayBuffer (aka one that was created in native) gets passed to JS, and then back from JS to native, the information that it is a native ArrayBuffer gets **lost** because we don't have `.isMutableBuffer()` and `.getMutableBuffer()` in JSI.
I created a feature request for this here: https://github.com/facebook/hermes/issues/1578 - @tmikov said he thinks it's a good idea - I even created a PoC PR that implements this (https://github.com/facebook/hermes/pull/1733), but it's not fully finished - might need a little bit of help here from the Hermes team.

In Nitro, I have a workaround for this by simply attaching a `NativeState` of the `jsi::MutableBuffer` pointer to my object (https://github.com/mrousavy/nitro/blob/cffffcb91f90fe4823fa7410bdb39c51c4c99125/packages/react-native-nitro-modules/cpp/jsi/JSIConverter%2BArrayBuffer.hpp#L81-L83) - works very well 😄 

But; this means it would not detect owning ArrayBuffers that were created natively in TurboModules, since TurboModules will likely not use `nitro::MutableBufferNativeState` lol. Not a biggie, it'll just be non-owning.
We need the Hermes PR for native support for that.

proposals/0000-arraybuffer-support-in-turbomodules.md
mrousavy: Agreed - ownership can transfer from native to JS here, that's perfectly fine. We use this in a lot of Nitro libraries.

proposals/0000-arraybuffer-support-in-turbomodules.md
mrousavy:
```suggestion
When passing an `ArrayBuffer` that was created in JS to native code, it should always be treated as "borrowed" or "non-owning": JS owns the ArrayBuffer's memory and the JS GC is responsible for freeing it. Native code should access the passed memory only for the duration of the synchronous call.
```

Let's clarify this a bit here; only if it was created in JS, we don't have a `std::shared_ptr<jsi::MutableBuffer>`.

If it was created in native, we can technically unwrap the native buffer again and can assume ownership safely. But; see my other discussion for more info on this

proposals/0000-arraybuffer-support-in-turbomodules.md
paradowstack: I unnecessarily coupled this RFC with the first PR in my head. Asynchronous functions are mentioned a bit later, but they should be outlined here as well as this is the final API we would like to have.

proposals/0000-arraybuffer-support-in-turbomodules.md
paradowstack: Great catch, will add a Blob Manager example to the next version of the RFC.

proposals/0000-arraybuffer-support-in-turbomodules.md
paradowstack: Thanks for the suggestion - will expand this section accordingly.

proposals/0000-arraybuffer-support-in-turbomodules.md
paradowstack: Good idea - will do that!

proposals/0000-arraybuffer-support-in-turbomodules.md
paradowstack: Missed that one - thanks!

proposals/0000-arraybuffer-support-in-turbomodules.md
paradowstack: I thought that aligning the ownership strategy for both directions would make sense - but of course you are right, it would work fine for case of "native to JS" direction.

proposals/0000-arraybuffer-support-in-turbomodules.md
paradowstack: I sure did read your documentation and source code before starting working on this 😄

paradowstack: Thanks for clarification!

proposals/0000-arraybuffer-support-in-turbomodules.md
mrousavy: Yup - both react-native-mmkv ([see `MMKVManagedBuffer.h`](https://github.com/mrousavy/react-native-mmkv/blob/v3.3.3/package/cpp/MMKVManagedBuffer.h)) and react-native-vision-camera ([see `SharedArray.h`](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/ios/FrameProcessors/SharedArray.h)) use raw JSI access to expose `ArrayBuffer` data. In VisionCamera I kinda overcomplicated it for a while with TypedArray support (that's primarily for `Uint8Array`), but still - it's a lot of effort to get right, and even tho I added a ton of code for that, it still has it's bugs. Like thread-safety.
Also; react-native-fast-tflite ([see `TypedArray.h`](https://github.com/mrousavy/react-native-fast-tflite/blob/main/cpp/jsi/TypedArray.h)) and vision-camera-resize-plugin ([see `ResizePlugin.mm`](https://github.com/mrousavy/vision-camera-resize-plugin/blob/3fde4714c0f72037fb4ff20a66a50701b0432a8d/ios/ResizePlugin.mm#L609-L610)) are using raw JSI ArrayBuffers.

In Nitro, i finally solved all of that - so MMKV benefits from that ✨ for free ✨, but VisionCamera is not on Nitro yet.
I can definitely understand the Motivation here, lol.

paradowstack:
> I agree with the concerns around thread-safety expressed in this thread here. Is there any prior art we can reference?

AFAIK other implementations (such as Nitro Modules or Expo Modules) has similar approach - ownership is not transferred from JS to Native, but borrowed. From what I see thread safety is either manual or ensured by copying the content (e.g. Expo Blob) - I am not aware of the other solution to this problem - happy to be corrected. 
 
> Is the operation model similar to [Single-ownership model where buffers are either borrowed (JS→Native, sync only) or transferred (Native→JS)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer)? Should we consider [Atomics](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics) as a complementary but necessary capability here?

The proposed operational model isn't similar to `SharedArrayBuffer` I would say - there is no synchronisation. It's more single-ownership model where buffers are either borrowed (JS→Native, sync only) or transferred (Native→JS). Can we somehow provide using JSI similar mechanism for safe concurrent access as these primitives offers? 

> Making this fully support async JS to native invocation calls will likely increase complexity, as it would require us to keep the JS object alive for the duration of the native memory reference, but not impossible. Alternatively, we'd need to make it really obvious that ArrayBuffer args can only be used in sync calls through codegen.

Yes, I can see we can do it either way. Would it also be a applicable solution to the "borrowed" vs "moved" discussion? Instead it can be "shared", by keeping JS object alive. It would not eliminate the thread-safety problems, but perhaps could be another solution to the problem.

paradowstack:
Thank you all for the feedback and discussion on this RFC! I've updated the proposal.

There are two open questions and fundamental design decisions that need input:

1. **Transfer vs. Borrowing Semantics** 
The current proposal uses borrowing semantics for JS→Native due to JSI/Hermes API limitations. However, as pointed out, this deviates from web standards and places unusual thread-safety burden on developers. Transfer semantics would solve this but from my understanding requires new JSI/Hermes APIs. Should we proceed with borrowing as a pragmatic interim solution? Do you see the other feasible solutions to the problem?
2. **Thread-Safety Model** 
Related to the above, we need to decide whether to require manual developer coordination or enforce thread-safety through ownership transfer or some synchronisation mechanism. 

I'd appreciate guidance from the core team and community on how to proceed with these questions!

paradowstack:
@javache did you maybe have time to take a look again at this RFC and open questions once again? 😊

tmikov:
Here are my general thoughts on the matter, and unfortunately I see a couple of fundamental problems (sorry, I should have joined here earlier):

1. ArrayBuffers are simply not very suitable for passing to asynchronous native code because they can be detached or resized. AFAICT, NodeJS doesn't have any asynchronous APIs that operate on ArrayBuffer precisely for that reason.
2. ArrayBuffers are also not suitable because technically there is nothing preventing the engine from allocating the buffer contents in the GC heap, thus making it **movable**. 
3. Using "transfer" semantics when calling native code would be unusual - after all we are not sending the array to another process or runtime. Again, NodeJS doesn't do that. Designing APIs around unusual techniques is questionable.
4. The threading issue is not a concern. Usually it is not the APIs job to protect a buffer from simultaneous access.

Problem 1 has partial workarounds - you can check whether the buffer is resizable and reject it, but you can't prevent it from becoming asynchronously detached. Problem 2 is basically unsolvable. 

This is why NodeJS uses Buffer - it can fully control it.

My recommendation: give up on ArrayBuffer. It simply can never work well because of problem 2. It will always be somewhat of a hack. Use Buffer instead, which is well understood by everybody, compatible with NodeJS, etc, etc.

If you folks are dead-set on ArrayBuffer, I think @mrousavy's proposal for `getMutableBuffer()` is a solution. If the ArrayBuffer was created by native code as a `MutableBuffer`, then obtaining a `shared_ptr` to it gives sufficient guarantees, I think. If it wasn't, it should be rejected at runtime by the API. Personally I don't think that would be a great API, but we are going to add `getMutableBuffer()` anyway.

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,499 @@
...
+
+Developers should be able to define a TypeScript spec as follows:
+
+```ts
+export interface Spec extends TurboModule {
+  getBuffer(): ArrayBuffer;
+  processBuffer(buffer: ArrayBuffer): void;
+
+  getAsyncBuffer(): Promise<ArrayBuffer>;
+  processAsyncBuffer(buffer: ArrayBuffer): Promise<void>;
```
javache, line 26: Let's remove support for this until we have `getMutableBuffer` and can provide a safe variant of this.

proposals/0000-arraybuffer-support-in-turbomodules.md
```diff
@@ -0,0 +1,499 @@
...
+
+Until these underlying platform capabilities are available, borrowing semantics provide a pragmatic path forward while acknowledging the thread-safety implications documented in the Threading section below.
+
+**Alternatives**: Alternatively, transfer of the ownership to the native can be mimicked by aligning the JS object lifetime with native memory reference. It would still allow access to the buffer on the JS side, potentially resulting in race conditions. However, it can be clearly stated in the documentation that once a buffer is passed to Native, it cannot be used anymore.
+
+#### Native-to-JS (Transfer Ownership)
+
+When passing a buffer from native to JS, ownership can be transferred, and it can be treated as "owned". Native code creates the buffer, wraps it in a shared pointer, and passes it to JS. The JS GC will properly manage the buffer's lifetime, and the buffer will be freed when no longer referenced.
```
javache, line 70: In this scenario (of a shared_ptr), ownership is shared through ref counting, not transferred.

javache (approved):
I'm ok with proceeding here with the constraints @tmikov points out - let's enforce that when we bridge ArrayBuffers originating from JS that these are marked as unmovable (and not supported in async calls) and let's pull in https://github.com/facebook/hermes/pull/1733 to avoid copying data when async scenarios are needed in the native layer.

paradowstack:
> I'm ok with proceeding here with the constraints @tmikov points out - let's enforce that when we bridge ArrayBuffers originating from JS that these are marked as unmovable (and not supported in async calls) and let's pull in [facebook/hermes#1733](https://github.com/facebook/hermes/pull/1733) to avoid copying data when async scenarios are needed in the native layer.

Hi @javache, great to hear you're okay with proceeding with the constraints. 
I’m currently tied up with other topics, but when I return to this, I’ll make some small adjustments to the RFC to get it into its final shape.

PR #967, opened 2025-12-16
iOS SceneDelegate migration proposal

artus9033:
A proposal for adding `SceneDelegate` support to iOS.

[View the rendered RFC](https://github.com/artus9033/discussions-and-proposals/blob/artus9033/scenedelegate/proposals/0000-ios-scenedelegate-migration.md)


proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+![](assets/0000-ios-deprecation-warning.png)
+
+It is possible to perform dynamic window resizing both via stage manager (split view) and freely via the bottom-right corner handle. `SceneDelegate` with proper configuration in `Info.plist` can enable the multi-window capability for iPadOS, which would require adjustments in RN code and RN libraries code to accommodate such design. Such a change would be large and shall be addressed in a separate RFC. The scope of this RFC is to cover a single resizable window (single instance of a given RN app), yielding the assumption that `UIApplicationSupportsMultipleScenes` **must not** be set to true at the moment. This RFC will bring in an additive change that should allow for adoption of a new `SceneDelegate` entrypoint while also allowing for (deprecated) usage of `AppDelegate`. Also, the `@react-native-community/template` and `packages/rn-tester` should be migrated to implement the `SceneDelegate` API.
```
cipolleschi:
```suggestion
This RFC will bring in an additive change that should allow for adoption of a new `SceneDelegate` entrypoint while also allowing for (deprecated) usage of `AppDelegate`. Also, the `@react-native-community/template` and `packages/rn-tester` should be migrated to implement the `SceneDelegate` API.

:::note
It is possible to perform dynamic window resizing both via stage manager (split view) and freely via the bottom-right corner handle. `SceneDelegate` with proper configuration in `Info.plist` can enable the multi-window capability for iPadOS, which would require adjustments in React Native code and React Native libraries code to accommodate such design. Such a change would be large and shall be addressed in a separate RFC. The scope of this RFC is to cover a single resizable window (single instance of a given React Native app), yielding the assumption that `UIApplicationSupportsMultipleScenes` **must not** be set to true at the moment. 
:::
```

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+One related aspect not covered by this RFC is the performance of `useWindowDimensions` hook, which during intensive resizing of the application behaves suboptimal. This matter shall be addressed in a separate RFC.
```
cipolleschi:
```suggestion
```

this is unrelated to the SceneDelegate migration

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
+---
```
cipolleschi: rename the file to `ios-scenedelegate-migration`

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+One related aspect not covered by this RFC is the performance of `useWindowDimensions` hook, which during intensive resizing of the application behaves suboptimal. This matter shall be addressed in a separate RFC.
+
+The idea proposed in this RFC is to:
+
+1) migrate from `AppDelegate` to `SceneDelegate`  
+2) educate on the need to migrate existing apps to adopt `SceneDelegate`
```
cipolleschi:
```suggestion
```

This can be removed. 1) is the topic of the RFC, and 2) is implicit: part of an RFC is educating the users.

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+@end
+```
+
+### Migration from `AppDelegate` to `SceneDelegate`
+
+Adoption of UIScene lifecycle requires the following actions:
+
+* In application base code  
+  * migration from `AppDelegate` as the primary point of lifecycle-related logic to `SceneDelegate`; for backwards compatibility, RN public API integration points will still be compatible with `AppDelegate` approach for users that may not want to migrate immediately  
+  * invoke RN `RCTLinkingManager` methods from `SceneDelegate`:
```
cipolleschi:
```suggestion
  * invoke React Native `RCTLinkingManager` methods from `SceneDelegate`:  
```

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+@end
+```
+
+### Migration from `AppDelegate` to `SceneDelegate`
+
+Adoption of UIScene lifecycle requires the following actions:
+
+* In application base code  
+  * migration from `AppDelegate` as the primary point of lifecycle-related logic to `SceneDelegate`; for backwards compatibility, RN public API integration points will still be compatible with `AppDelegate` approach for users that may not want to migrate immediately
```
cipolleschi:
```suggestion
  * migration from `AppDelegate` as the primary point of lifecycle-related logic to `SceneDelegate`; for backwards compatibility, React Native public API integration points will still be compatible with `AppDelegate` approach for users that may not want to migrate immediately  
```

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+* In application base code  
+  * migration from `AppDelegate` as the primary point of lifecycle-related logic to `SceneDelegate`; for backwards compatibility, RN public API integration points will still be compatible with `AppDelegate` approach for users that may not want to migrate immediately  
+  * invoke RN `RCTLinkingManager` methods from `SceneDelegate`:  
+    * `scene:continueUserActivity:`
+    * `scene:openURLContexts:`
+  * update of the app's Info.plist to include a UIApplicationSceneManifest specifying the support and disabling multiple scenes capability  
+* In React Native code:  
+  * migration of code that relies on `launchOptions` and deprecated `UIApplicationLaunchOptions`\* keys to UIScene lifecycle and UIScene.ConnectionOptions.userActivities  
+* In React Native code, RN native libraries’ code:
```
cipolleschi:
```suggestion
* In React Native code, React Native native libraries’ code:  
```

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+- [application:openURL:options:](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/application\(_:open:options:\)?language=objc) in favor of [scene(\_:openURLContexts:)](https://developer.apple.com/documentation/uikit/ui`scenedelegate`/scene\(_:openurlcontexts:\)) \- used by RCTLinkingManager  
+- non-UIScene lifecycle: [https://developer.apple.com/documentation/technotes/tn3187-migrating-to-the-uikit-scene-based-life-cycle](https://developer.apple.com/documentation/technotes/tn3187-migrating-to-the-uikit-scene-based-life-cycle) \- used by base application code
+
+## Detailed design
+
+### `SceneDelegate` `RCTReactNativeFactory` entrypoint
+
+To support `SceneDelegate` lifecycle, a new entrypoint for RN application initialization should be provided in `RCTReactNativeFactory`, which would be invoked from `SceneDelegate` lifecycle methods. The existing entrypoint from `AppDelegate` would be kept for backwards compatibility, making this an additive change. An example code snippet of a ``SceneDelegate`` using React Native proposed in this RFC would be:
```
cipolleschi:
```suggestion
To support `SceneDelegate` lifecycle, a new entrypoint for React Native application initialization should be provided in `RCTReactNativeFactory`, which would be invoked from `SceneDelegate` lifecycle methods. The existing entrypoint from `AppDelegate` would be kept for backwards compatibility, making this an additive change. An example code snippet of a ``SceneDelegate`` using React Native proposed in this RFC would be:
```

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+  * migration of code that relies on `launchOptions` and deprecated `UIApplicationLaunchOptions`\* keys to UIScene lifecycle and UIScene.ConnectionOptions.userActivities  
+* In React Native code, RN native libraries’ code:  
+  * migration of app lifecycle methods from application\* to scene\* as per [https://developer.apple.com/documentation/technotes/tn3187-migrating-to-the-uikit-scene-based-life-cycle](https://developer.apple.com/documentation/technotes/tn3187-migrating-to-the-uikit-scene-based-life-cycle)  
+  * migration of code that relies on *any* *deprecated* `AppDelegate`-related APIs, which detailed description of is presented below
+
+To enforce the user not to enable the `UIApplicationSceneManifest`.`UIApplicationSupportsMultipleScenes` capability in `Info.plist`, from now on, the following warning would be printed from `RCTReactNativeFactory` provided the user had enabled the multi-window capability:
+
+![](assets/0000-console-warning-unsupported-configuration.png)
+
+#### Migration: RN app template, RNTester
```
cipolleschi:
```suggestion
#### Migration: React Native app template, RNTester
```

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+In case of the RN app template, the iOS boilerplate code is limited only to basic bootstrapping of the application. This implies adjustments to:
```
cipolleschi:
```suggestion
In case of the React Native app template, the iOS boilerplate code is limited only to basic bootstrapping of the application. This implies adjustments to:
```

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+#### Migration: `RCTLinkingManager`
+
+The linking manager is using ``AppDelegate`` methods for handling URLs being opened at runtime. This needs to be migrated to `SceneDelegate`. To maintain backwards compatibility, we can implement both approaches and \- to ensure that only one is invoked at a given time \- conditionally check if the app is based on ``AppDelegate`` or has scenes to ensure only one listener handles the event. The Scene lifecycle options (`NSDictionary`) are adapted to the format of ``AppDelegate`` launchOptions (`NSDictionary` as well).
+
+#### Migration: RCTDevLoadingView
+
+The does not account for updating the overlay `UIWindow` constraints so it does not get resized along with the main window. The proposed fix is to update the constraints based on KVO observation of the main window’s frame.
+
+![](assets/0000-RCTDevLoadingView-width.png)
+![](assets/0000-RCTDevLoadingView-width-clipped.png)
```
cipolleschi: this part might have changed a little in the meantime. Worth rechecking the status of the codebase.

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### Migrating Dimensions: internal state adjustment & naming convention
+
+Dimensions currently are exported as a constant. Such design does not fit the concept of resizable windows. To accommodate this, `RCTDeviceInfo` (iOS native module) should update its internal state variable upon changes to the frame so as to make `getConstants` return a value that is up-to-date at the time of invocation.
+
+`Dimensions` (JS API) contains a `getConstants` method that wraps the native `getConstants` method and caches the result internally. The caching would now be obsolete (since the value may change in time), therefore the underlying native method should always be invoked. Moreover, the naming of the JS `getConstants` method in face of resizable windows may be misleading, since from now on the dimensions are not constant. Therefore, I propose a gradual adoption of a new method of same functionality, `getInfo()`, along with the deprecation of `getConstants()`, on the JS API side. A rough draft of the JS API changes would be:  
+
+![](assets/0000-nativedeviceinfo-api.png)
```
cipolleschi: I don't think this is actually required for the SceneDelegate migration. probably we can move it to a different RFC/effort

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+  - Ones not using native code:  
+    - adjusting `Info.plist` and the entrypoint according to the upgrade helper diff  
+    - ensuring that consumed libraries work well in multi-instance setups   
+  - Ones using native code:  
+    - the above  
+    - migrating from `AppDelegate` lifecycle methods to `UIScene` lifecycle methods  
+    - inspecting usage of singletons & static fields to ensure they logically fit the multi-instance reality  
+- In RN libraries:  
+  - migrating from `AppDelegate` lifecycle methods to `UIScene` lifecycle methods  
+  - inspecting usage of singletons & static fields to ensure they logically fit the multi-instance reality
```
cipolleschi: These are not really drawbacks... I'd love to see this section reworded as an action plan, where each item is a task (`- [ ] Task 1`) so that we can track the implementation.

The only drawback of this RFC is hat users will have to migrate and we have to handle an ecosystem migration.

proposals/0000-ios-scenedelegate-and-resizing.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Alternatives
+What other designs have been considered? Why did you select your approach?
```
cipolleschi:
```suggestion
No alternatives have been considered. Apple will enforce the change and we have to comply with it.

An alternative way to approach it is to implement some breaking changes without preserving backward compatibility for a while. This can backfire as our users could churn away from React Native.
```

cipolleschi (changes requested):
Good job here! 
I left a bunch of comments that should be applied/adressed in the RFC.

Make sure not to have any `RN` acronym. In official documentation we always use the full React Native product name.

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: Sure, I will validate this on the newest main

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: Agreed, I will move it from here, then

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: Absolutely, that makes perfect sense

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: Thank you for this! I was not sure what to do with this section, but this is perfect for everyone's context

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: It seems like a dismiss button has been added, but that doesn't collide - anyway, I will precise a better idea to accomodate the resizing behaviour. Thanks for the heads-up!

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: Thanks!

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: Got it, removed now

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: Agreed, thank you!

artus9033:
Thank you for the comments Riccardo! Yes, sorry for not using the full name of React Native - I will keep that in mind. I think the RFC is now ready for a re-review, I applied your suggestions and also reworded few sections.

proposals/0000-ios-scenedelegate-migration.md
```diff
@@ -0,0 +1,159 @@
...
+
+@end
+```
+
+### Migration from `AppDelegate` to `SceneDelegate`
+
+Adoption of `UIScene` lifecycle requires the following actions:
+
+* In application base code  
+  * migration from `AppDelegate` as the primary point of lifecycle-related logic to `SceneDelegate`; for backwards compatibility, React Native's public API integration points will still be compatible with `AppDelegate` approach for users that may not want to migrate immediately
```
TheRogue76, line 93: Think it would be ok to not have that backwards compatibility. The Delegate is on a clock anyway that we do not control, so everyone NEEDS to migrate, and keeping the old approach around is just delaying an inevitable that will be significantly more painful then and we would need to live the double life in the mean time supporting both approaches.
Might be best to just say, from this version of RN, in order to be compliant with Apple's requirements, we are moving off, older versions not affected. Just being clear in the comms with the framework end users

artus9033: I agree on the point that this has to be removed at some stage, yet the objective I had in mind was to bring an additive, non-breaking change. We should definitely deprecate those methods, but a sudden removal would make an unpleasant surprise for the users. By deprecating, we make it a gentle migration and we'll get to removal anyway.

proposals/0000-ios-scenedelegate-and-resizing.md
artus9033: I opened a separate RFC to track this idea: https://github.com/react-native-community/discussions-and-proposals/pull/969 - CC @cipolleschi

proposals/0000-ios-scenedelegate-migration.md
matinzd: I agree with Parsa. It is better to do this properly once rather than carrying things forward just for backward compatibility. Many things in the core break over time without being visible to us, and this is not an exception.

Another issue is the terminology itself. In the React Native ecosystem, the term “users” is often used to refer to consumers of a package, which easily causes confusion. In our case, those “users” are developers, not end users. Since the audience is developers, they are expected to adapt to changes and learn new patterns. We are not building a product for non-technical end customers.

artus9033: Thanks for sharing that, then I think the matter is up to be discussed. WDYT @cipolleschi ?

Regarding the "users", definitely the user persona I was referring to in this context was the developer - yet, I'll keep in mind that the default meaning here is the end user, thank you for noting that!

objectiveSee:
Hey, this is great work. I'm exploring CarPlay support in a React Native/Expo app where CarPlay would be implemented natively in Swift using CPTemplate APIs, while the main app stays in React Native.

A few questions about the proposed API:

1. Will RCTReactNativeFactory gracefully handle scenes it doesn't manage? In my case, the CarPlay scene would be pure Swift with no React Native involvement.

2. Should the factory be instantiated globally or only for React Native-managed scenes?

3. Any caveats for scene role routing when some scenes bypass React Native entirely?

artus9033:
> Hey, this is great work. I'm exploring CarPlay support in a React Native/Expo app where CarPlay would be implemented natively in Swift using CPTemplate APIs, while the main app stays in React Native.
> 
> A few questions about the proposed API:
> 
Thank you, it's a pleasure to hear that! Regarding your questions:

> 1. Will RCTReactNativeFactory gracefully handle scenes it doesn't manage? In my case, the CarPlay scene would be pure Swift with no React Native involvement.

I think it should not collide in any way, since scene management happens in the driver code of your application:

```objc

- (void)scene:(UIScene *)scene
    willConnectToSession:(UISceneSession *)session
                 options:(UISceneConnectionOptions *)connectionOptions
{
  // ...
  RCTReactNativeFactory *factory = [[RCTReactNativeFactory alloc] initWithDelegate:delegate];
  // ...
  [factory startReactNativeWithModuleName:@"RNTesterApp"
                                 inWindow:self.window
                        initialProperties:[self prepareInitialProps]
                        connectionOptions:connectionOptions];
}
```

That said, React Native operates only in the window in the scene you start it with.

> 2. Should the factory be instantiated globally or only for React Native-managed scenes?

I think this is a question more related to brownfield, this PR does not change anything w.r.t. the previous state, but anyway the factory should be initialized once from what I remember - as in [this piece of code](https://github.com/callstack/react-native-brownfield/blob/a4da8e1965d652c351c9782c0b9c260552a20996/packages/react-native-brownfield/ios/ReactNativeBrownfield.swift#L114-L117) in react-native-brownfield.

> 3. Any caveats for scene role routing when some scenes bypass React Native entirely?

Just to confirm, by that you mean creating some scene roles that never have anything to do with React Native? I haven't done any experiments in such a scenario, but don't foresee anything that would break - the code path for initializing the RN scene would just never be hit, I guess.

matinzd:
When are we thinking of implementing this. I guess the RFC is pretty clear and good to go.

artus9033:
Hi @matinzd ! I raised a PR here: https://github.com/react/react-native/pull/53744
I hope we will be able to merge it in the upcoming weeks.

PR #994, opened 2026-04-21
RFC: (iOS) Swift Package Manager in React Native (Cocoapods deprecation)

chrfalch:
This RFC contains a plan and technical description for how to deprecate Cocoapods and replace it with Swift Package Manager on iOS.

[RFC Document](https://github.com/chrfalch/discussions-and-proposals/blob/chrfalch/add-rfc-draft-for-spm/proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md)

### DIFF
--- a/proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
+++ b/proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
@@ -86,7 +86,7 @@ injects the SPM package references into the app's existing Xcode project, in pla
 
 From the app’s point of view, the integration stays relatively simple: the app's Xcode project references three locally generated packages — React Native core, generated code, and autolinked native modules — through `XCLocalSwiftPackageReference` entries. The injection also adds an auto-sync build phase that refreshes autolinking when dependencies change. No app-level `Package.swift` is generated or required; the sub-package paths under `build/` are stable, so adding or removing community deps never requires re-injecting. The edit is purely additive and recorded in a `.spm-injected.json` marker, so it can be reversed exactly (`deinit`) without touching the rest of the project.
 
-This RFC does not propose a new packaging model from scratch. Instead, it proposes standardizing and evolving the implementation direction already demonstrated in the prototype branch.
+This RFC does not propose a new packaging model from scratch. Instead, it proposes standardizing and evolving the implementation direction already demonstrated in the prototype branch. The tooling, commands, header model, and extension points described below are documented in the branch under `packages/react-native/scripts/spm/__doc__/` (start with `spm-scripts.md`).
 
 ### Autolinking
 
@@ -165,12 +165,16 @@ Instead of asking each app to compile this dependency graph from source, the SPM
 
 These artifacts are downloaded during setup and cached locally, and the generated SPM package for the app links against them as binary targets. This keeps the app-facing package graph small and avoids moving the full complexity of the React Native native dependency tree into every application project.
 
-At the implementation level, the branch also addresses one of the key technical challenges of using prebuilt artifacts with React Native: header compatibility. React Native uses both framework-style includes such as `#import <React/...>` and many lowercase namespace includes such as `#import <react/...>`, `<yoga/...>`, and `<jsi/...>`, which SPM does not handle directly. The prototype solves this with two artifacts shipped together:
+At the implementation level, the branch also addresses one of the key technical challenges of using prebuilt artifacts with React Native: header compatibility. React Native uses framework-style includes such as `#import <React/...>`, many lowercase namespace includes such as `#import <react/...>`, `<yoga/...>`, and `<jsi/...>`, and the third-party dependency namespaces (`folly/`, `glog/`, `boost/`, …) — none of which SPM serves directly from a plain framework binary. The prototype solves this with a set of artifacts shipped together:
 
 - `React.xcframework`, whose framework module map serves the `#import <React/...>` headers
-- `ReactNativeHeaders.xcframework`, a headers-only artifact whose module map makes the lowercase namespaces (`react/`, `yoga/`, `jsi/`, `ReactCommon/`, …) modular, so those includes resolve through standard Clang module lookup
+- `ReactNativeHeaders.xcframework`, a headers-only artifact whose module map makes React Native's own lowercase namespaces (`react/`, `yoga/`, `jsi/`, `ReactCommon/`, …) modular
+- `ReactNativeDependencies.xcframework`, the prebuilt third-party dependency binaries, paired with `ReactNativeDependenciesHeaders.xcframework`, a headers-only sidecar that serves the dependency namespaces (`folly/`, `glog/`, `boost/`, …). The binary is a framework and cannot expose those headers to SwiftPM, so the headers ship as their own library-type artifact — which also keeps `ReactNativeHeaders` purely React Native, giving every namespace exactly one home
+- `hermes-engine`, the Hermes runtime
 
-(An earlier iteration used a generated Clang Virtual File System overlay for the non-standard includes; that approach was replaced by the `ReactNativeHeaders` module map, which needs no per-app overlay generation.)
+so those includes resolve through standard header search paths and Clang module lookup with no per-app configuration.
+
+(An earlier iteration used a generated Clang Virtual File System overlay for the non-standard includes; that approach was replaced by the headers-only artifacts above, which need no per-app overlay generation.)
 
 This means the RFC does not need to argue only in theory that Hermes and React Native dependencies can work under SPM. The prototype already demonstrates a concrete model based on prebuilt XCFrameworks and local package generation.
 
@@ -234,6 +238,14 @@ During the transition period, libraries should be able to ship both:
 - a `podspec` for CocoaPods users
 - SPM metadata and, optionally, `Package.swift` or prebuilt XCFramework artifacts for SPM users
 
+### Framework plugins (autolinking hooks)
+
+Some ecosystems — Expo is the first — layer their own module system on top of React Native: they discover native modules dynamically, generate a module registry, and ship mixed Swift/Objective-C/C++ modules such as `ExpoModulesCore`. A static `spm.modules` list cannot express this, and a one-shot post-process of the generated `Package.swift` would be overwritten the next time the auto-sync build phase re-runs autolinking.
+
+The prototype therefore exposes a generic **autolinking plugin hook**. Any dependency can register a plugin from its own `react-native.config.js` (a `spm.autolinkingPlugin` entry), and React Native invokes it wherever autolinking runs — during `add`/`update` and during the build-time sync — so a framework's contribution is regenerated on every build and never goes stale. The plugin receives the app/JS-root context and the discovered autolinking metadata, and returns SwiftPM package references, product dependencies, and generated-source files that React Native merges into the autolinked package graph. Discovery is transitive — installing the framework is enough, mirroring how CocoaPods pulls in `use_expo_modules!` — with an app-level opt-out for excluding a plugin.
+
+This keeps React Native framework-agnostic (no framework-specific code in core) while giving frameworks the same extension seam CocoaPods provided through the Podfile, `use_expo_modules!`, and its post-install hooks. Expo is the first consumer and is validating the contract, which ships as an unstable preview until proven. The discovery mechanism and the plugin contract are documented in `packages/react-native/scripts/spm/__doc__/spm-autolinking-plugins.md`.
+
 ### Transition phases
 
 The migration should happen in stages:
@@ -246,7 +258,7 @@ The migration should happen in stages:
 
 ## Brownfield
 
-Brownfield projects will also be able to use Package.swift files and integrate easily with existing Swift or Objective-C based apps, either as precompiled XCFrameworks or as source - both defined in a Package.swift file. There will be an option in the cli for generating separate Package.swift files for distributing the app as a brownfield target.
+Because `add` injects into the app's existing Xcode project in place (rather than generating one), an app that embeds React Native can adopt SPM today by pointing `add` at the right project and target (`--xcodeproj` / `--product-name`); the current constraint and coexistence rules are described in the tool docs (`spm-scripts.md`, "Brownfield apps"). Building on that, brownfield projects will also be able to use Package.swift files and integrate easily with existing Swift or Objective-C based apps, either as precompiled XCFrameworks or as source — both defined in a Package.swift file. There will be an option in the cli for generating separate Package.swift files for distributing the app as a brownfield target.
 
 ## Drawbacks
 
--- a/proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
+++ b/proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
@@ -28,33 +28,37 @@ If React Native intends to deprecate CocoaPods in December 2026, it needs a migr
 
 NOTE: The following examples use the demo branch but will be implemented in react-native-cli with the same parameters and interface.
 
-For a new project, the prototype SPM setup is initialized with:
+For a new project, SPM is set up with:
 
 ```bash
-yarn ios --init # it calls node node_modules/react-native/scripts/setup-ios-spm.js --init under the hoods
-
+npx react-native spm  # invokes node node_modules/react-native/scripts/setup-apple-spm.js
+                      # with no action — auto-resolves to the right verb (see below)
 ```
 
-That command:
+The command exposes four verbs; with no action it auto-resolves which one to run:
 
-- runs React Native Codegen,
-- generates `autolinked/Package.swift`,
-- downloads the required prebuilt XCFramework artifacts,
-- generates the local SPM packages consumed by the app, and
-- creates an Xcode project configured for the SPM flow.
+- `add` — runs React Native Codegen, generates
+  `build/generated/autolinking/Package.swift`, downloads the prebuilt XCFramework
+  artifacts, generates the local SPM packages, and **injects** the SPM package
+  references into the app's existing Xcode project, in place. The injection is
+  additive and idempotent, and is recorded in a `.spm-injected.json` marker.
+  Default on a first run.
+- `update` — re-runs the pipeline and refreshes the existing injection. Default
+  once a project has been set up.
+- `deinit` — the exact inverse of `add`: surgically removes only what `add`
+  injected (using the marker) and drops it. Git-recoverable; no prompt.
+- `scaffold` — generates a `Package.swift` for community libraries that ship only
+  a podspec.
 
-After initialization, manual refresh is done with:
-
-```bash
-yarn ios # it calls node node_modules/react-native/scripts/setup-ios-spm.js under the hoods
-```
-
-And a full reset of generated state is done with:
-
-```bash
-yarn ios --clean # it calls node node_modules/react-native/scripts/setup-ios-spm.js --clean under the hoods
-
-```
+For an app that still uses CocoaPods, `add --deintegrate` first runs
+`pod deintegrate` and strips React Native from the Podfile, then injects SPM —
+converting the project in one step. The zero-arg `npx react-native spm` detects a
+freshly-scaffolded CocoaPods project and does this automatically; on an
+established app it stops and points the developer at `--deintegrate`. To remove
+SPM again, `npx react-native spm deinit` (then `pod install` to return to
+CocoaPods). Because the integration is injected in place rather than generated
+from scratch, the app keeps its own `.xcodeproj` (signing, capabilities, build
+phases) — no project is renamed or replaced.
 
 ## Detailed design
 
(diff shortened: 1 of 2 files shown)


proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+
+- organized headers inside the XCFramework artifacts
+- a generated Clang Virtual File System (VFS) overlay for the remaining non-standard include cases
+
+This means the RFC does not need to argue only in theory that Hermes and React Native dependencies can work under SPM. The prototype already demonstrates a concrete model based on prebuilt XCFrameworks and local package generation.
+
+### Build from source
+
+Build React Native from source will be supported in the final version of the SPM package system, but the current POC does not support it.
+
```
mrousavy, line 184:
```suggestion

### Mixing C++ and Swift

As per [the "Setting Up Mixed-Language Swift and C++ Projects" Swift documentation](https://www.swift.org/documentation/cxx-interop/project-build-setup/#mixing-swift-and-c-using-swift-package-manager), Swift Package Manager (SPM) **does not yet support Swift and C++ in the same target.**

This causes issues for Swift + C++ libraries or frameworks, including but not limited to the entire [Nitro Modules](https://nitro.margelo.com) ecosystem, as Nitro is built ontop of the direct Swift + C++ interop feature, and uses calls from Swift -> C++, as well as calls from C++ -> Swift within the same target/library - aka bi-directional interop.

This should not impose a problem for react-native directly as react-native uses Objective-C(++) for interoping between the languages (even Swift - C++ interop uses Objective-C(++) as a bridge, not **direct** Swift C++ interop).

The [Margelo](https://margelo.com) team is investigating whether this can be solved easily in Nitro - e.g. by using separate targets (one C++, one Swift, and possibly a third one to somehow break the bi-directional dependency).

As of today, this will affect popular libraries like react-native-vision-camera, react-native-mmkv, react-native-unistyles, react-native-video, and more.

```

I'll make sure to find a solution for this soon. 

Image

mrousavy (approved):
Overall this is a really exciting! 
I added a section about Mixing C++ and Swift with the direct interop, used by Nitro and all Nitro libraries as this is quite relevant to the ecosystem imo.
I'm sure I can find a solution for this soon.

Thanks for the great RFC Christian!

atlj (approved):

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+The current prototype already implements a working SPM packaging model in the chrfalch/swift-package-manager branch.
+
+At a high level, React Native core is exposed to app projects as a small set of Swift packages centered around prebuilt XCFrameworks:
+
+- `ReactNative`, which packages the React Native iOS runtime
+- `ReactNativeDependencies`, which packages the native third-party dependencies used by React Native
+- `hermes-engine`, which provides the Hermes runtime
+
+The prototype uses a single script, `setup-ios-spm.js`, to prepare an app for this layout. That script:
```
Saadnajmi: A slight nit: Since this will be used for iOS and Mac Catalyst in React Native Core, tvOS in react-native-tvos, and macOS + visionOS in React Native macOS, could we name this something like `setup-apple-spm.js`? There's a lot of places where `ios` is hardcoded (`build/generated/ios` the biggest one that comes to mind) and I think it's more representative to say this is for the apple ecosystem of platforms.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+At a high level, the flow is:
+
+1. React Native Codegen generates the native metadata and codegen artifacts for the app.
+2. The SPM autolinking step reads the generated autolinking metadata and the app’s `react-native.config.js`.
+3. It produces `autolinked/Package.swift`, which contains local package targets for the app’s native dependencies.
+4. The app’s generated Xcode project depends on that package, so native modules are available through normal SPM package resolution.
+
+This keeps the developer-facing model close to today’s React Native workflow: dependencies are still discovered from the JavaScript project, but the integration backend is SPM instead of CocoaPods.
+Local modules
+The prototype also supports additional local modules through `spmModules` in `react-native.config.js`. This makes it possible to include native code that is not otherwise discovered through standard autolinking.
```
Saadnajmi: One edge case that used to not work but I think does now is a monorepo with several library packages, and one app. Old autolinking used to not work and I would hardcode extra pods in my podfile that pointed to the "local" podspecs, but nowadays autolinking works with local packages too. It would be great if that kept working with SPM too.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+
+```bash
+yarn ios --init
+```
+
+This first-time setup generates the initial `Package.swift`, the generated SPM packages, and the Xcode project for the app.
+
+After initialization, autolinking is normally kept up to date automatically by the generated Xcode project. The prototype adds an auto-sync build phase that detects dependency changes and regenerates the autolinked package before compilation when needed.
+
+When a manual refresh is needed, for example after changing native dependencies or local module configuration, developers can re-run:
```
Saadnajmi: Is the idea that if someone adds a node module with native code, there's a script phase somewhere that detects that the user might be missing an autolinked pakage?
I.E:
```
yarn ios --init
yarn add @shopify/react-native-skia
# User runs and builds the app without running `yarn ios`
```
Would there be an error, or does the "auto-sync build phase" regenerate `autolinked/Package.Swift`?? I suppose this is similar to how RN devs know they need to `pod install` constantly, and much better than adding a post install.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+
+Third-party library support should follow the same general model as React Native core: libraries should be consumable through SPM, while still allowing an incremental migration period where CocoaPods and SPM coexist.
+
+The simplest way forward is to add support for building using SPM by adding a Package.swift file to the library.
+
+The current prototype already supports app-local native modules through `spmModules` in `react-native.config.js`. Each declared module is turned into a target in `autolinked/Package.swift`, with sources mirrored into the generated package layout so they can be compiled by SPM.
+
+That mechanism is enough to prove that source-based native modules can be integrated into the SPM flow today.
+
+For third-party libraries published to npm, the longer-term direction should be to add SPM metadata alongside the existing React Native autolinking metadata in `react-native.config.js`. That keeps React Native’s dependency discovery in one place and lets the CLI/autolinking pipeline decide how to integrate each library.
```
Saadnajmi, line 195: To confirm, autolinking would not do what the CLI does today with cocoapods and look for Package.Swift files at the root of every node module? That seems fine, just wanted to confirm.

Saadnajmi (approved):
Thank you so much for this! I look forward to the new future. I left some minor comments and nits, but this looks very well thought out 👍

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+
+1. Experimental support: React Native ships opt-in SPM support for iOS.
+2. Dual support: CocoaPods and SPM are both officially supported.
+3. Default for new apps: once ecosystem coverage is good enough, SPM becomes the default for new iOS projects.
+4. Deprecation: CocoaPods is deprecated in December 2026.
+5. Removal of Cocoapods from the codebase: at some point after 12/2026. This depends on the rate of adoption by the community. We will keep it monitored as we did for the New Architecture rollout.
+
+## Brownfield
+
+Brownfield projects will also be able to use Package.swift files and integrate easily with existing Swift or Objective-C based apps, either as precompiled XCFrameworks or as source - both defined in a Package.swift file. There will be an option in the cli for generating separate Package.swift files for distributing the app as a brownfield target.
```
hurali97: >There will be an option in the cli for generating separate Package.swift files for distributing the app as a brownfield target.

Is this intended to support a similar use-case we previously did with `Podfile` using the following code?

```ruby
target 'BrownfieldLib' do
  inherit! :complete
end
```

Context: In brownfield apps, when we create a new framework target inside a RNApp, we add the above block in `RNApp/ios/Podfile` to inherit the package linking and all to the framework target.

Since this behavior may not directly be available with SPM - I am wondering whether the suggested approach of creating a separate Package.swift for the framework target will have duplications containing the same parts as App Target's Package.swift file.

hurali97 (approved):
Hey, thanks for the very well put RFC 🚀 

From Brownfield perspective, I have two points in addition to a review:

- https://github.com/react-native-community/discussions-and-proposals/pull/994#issuecomment-4311096830
- https://github.com/react-native-community/discussions-and-proposals/pull/994#issuecomment-4311097714

hurali97:
- static and dynamic frameworks

How are we planning to support this with SPM? With cocoapods, we would usually change this line to toggle between static and dynamic framework linkage:

```Podfile
linkage = ENV['USE_FRAMEWORKS']
```

Since with the final shape of SPM, we may not have this Podfile - so I wonder have you thought of any ideas on how we support it?

This in general should be for both greenfield and brownfield.

hurali97:
- Brownfield Packaging

This is rather third party library scoped https://github.com/callstack/react-native-brownfield. There we have a few brownfield packages and 3 of them are shipped as separate xcframeworks.

The user then have the flexibility to either:
- Drag and Drop them to their xcode projects
- Distribute them via SPM

We do not provide first class support for SPM but that is doable by the user. Now that this RFC is actually going to make that support in the core, we will also comply and extend it to our brownfield packaging. This does NOT in anyway means that the generated SPM from this package needs to be linked by the Brownfield Target - rather by the native host App consuming it.

One point on that:
If with this RFC, we ship a separate Package.swift for Brownfield Target, then we are thinking to opt for the following approach for packaging brownfield:

- Create a Package.swift with `ReactBrownfield.xcframework`, `brownfield-navigation.xcramework` and `brownie.xcframework` - as binaryTargets
- Reference the generated Package.swift for Brownfield Target inside this new Package.swift
- Add the binaryTargets to the Brownfield Target
- Ship this Package.swift to the native host App

The above is the layout of what I think we can do once we have this RFC delivered as a RC. Before that, we can also experiment with SPM and once RFC is released, we can change the local link for Brownfield Target to the Package.swift link and keep the rest Package.swift the same with binaryTargets.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+At the implementation level, the branch also addresses one of the key technical challenges of using prebuilt artifacts with React Native: header compatibility. React Native still uses many CocoaPods-style include paths such as `#import <React/...>`, which SPM does not handle directly. The prototype solves this by combining:
+
+- organized headers inside the XCFramework artifacts
+- a generated Clang Virtual File System (VFS) overlay for the remaining non-standard include cases
+
+This means the RFC does not need to argue only in theory that Hermes and React Native dependencies can work under SPM. The prototype already demonstrates a concrete model based on prebuilt XCFrameworks and local package generation.
+
+### Build from source
+
+Build React Native from source will be supported in the final version of the SPM package system, but the current POC does not support it.
```
tjzel, line 183: Does it mean that stepping into RN code while debugging will be only available when using Cocoapods, during the transition period?

tjzel (commented):
Thanks for the well-thought RFC, I like this direction.

I haven't had the opportunity to use SPM but I wonder if libraries can comfortably hook some scripts prior to compilation using it.

Libraries like Reanimated and Worklets currently execute some custom Ruby scripts inside `.podspec` to assert proper configuration and obtain compilation flags.

It would be perfect if library authors could extract their common logic from `build.gradle` and `.podspec` and put it into `react-native.config.js` file but I'm aware it's out of scope of this RFC.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
cipolleschi: Until we enable building from source with SPM, yes, you'll have to go back and use Cocoapods, unfortunately.

cipolleschi:
@tjzel I think you should be able to run [Swift PM Plugins](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/plugins/) to run prebuild code and ensure that everything is properly configured. That's a good point to add to the RFC, that libraries should be able to ship and configure Swift PM plugins. @chrfalch 

I believe it is ok if the first implementation doesn't have it, and we work with you folks from Reanimated to add the feature! 😉

chrfalch:
@tjzel and @cipolleschi: Build plugins have limited access to what they can do - but package resolving runs the Package.swift code and can contain almost anything you need, including updating script files or running separate pipelines. This is the equivalent of running pod install and will probably be the most popular hook for doing things like you mention.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: It doesn't mean that we will never support it - we just have to set a few limits to our 1 phase scope.

chrfalch:
> * static and dynamic frameworks
> 
> How are we planning to support this with SPM? With cocoapods, we would usually change this line to toggle between static and dynamic framework linkage:
> 
> ```
> linkage = ENV['USE_FRAMEWORKS']
> ```
> 
> Since with the final shape of SPM, we may not have this Podfile - so I wonder have you thought of any ideas on how we support it?
> 
> This in general should be for both greenfield and brownfield.

A package.swift file can include references to both prebuilt XCFrameworks and source files - so as long as the package you want to consume (like Firebase packages) is packaged as a framework you should be able to consume it from your package - without having to change anything in your project.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: @Saadnajmi - Not sure what you're asking for here? We will basically try to use the same logic for auto linking since it is agnostic when it comes to wether we're using Cocoapods or SPM - it is all about collecting the modules that needs to be linked.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: As for now the flow will basically be the same, when you update your dependencies you will need to update. This can of course be improved - the problem today is that running `pod install` is very slow - otherwise we'd be doing it automatically I guess.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: You will have a separate Package.swift files that you can commit and change as much as you like - it will reference another Package.swift file that contains the auto linked targets.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: Agree!

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+
+```
+
+That command:
+
+- runs React Native Codegen,
+- generates `autolinked/Package.swift`,
+- downloads the required prebuilt XCFramework artifacts,
+- generates the local SPM packages consumed by the app, and
+- creates an Xcode project configured for the SPM flow.
```
tido64: Is a new project created every time this script is called?

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+
+After initialization, manual refresh is done with:
+
+```bash
+yarn ios # it calls node node_modules/react-native/scripts/setup-ios-spm.js under the hoods
```
tido64: We use the `ios` "verb" to build the iOS project. Does this script also build the app or do we only call this script when we need to manually refresh the project?

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
tido64: Will there be any hooks for customizations that require running scripts?

chrfalch: No, only first time. You should be able to check in both the main Xcode project and Package.swift file.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: First version will most likely need you to run a script/command when updating dependencies like with Cocoapods.

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: Great idea, we haven't made any decisions about this yet - can you come up with some use cases for us to look into? (Remember that you will be able to run scripts from the resolve phase of your own package.swift code)

tido64: In [RNTA](https://github.com/microsoft/react-native-test-app/blob/trunk/packages/app/ios/test_app.rb), we also generate the Xcode project files when `pod install` is run. Additionally, we also:
- Disable warnings in third party libraries to reduce noise
- Define macros to enable/disable certain parts in third party libraries
- Add assets/resources declared in an external file
- Configure code signing (also declared in an external file)
- Apply config plugins (if applicable)
- Various scripts to validate configuration files
- Generate code that must be included in the project

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
```diff
@@ -0,0 +1,306 @@
...
+
+At a high level, the flow is:
+
+1. React Native Codegen generates the native metadata and codegen artifacts for the app.
+2. The SPM autolinking step reads the generated autolinking metadata and the app’s `react-native.config.js`.
+3. It produces `autolinked/Package.swift`, which contains local package targets for the app’s native dependencies.
+4. The app’s generated Xcode project depends on that package, so native modules are available through normal SPM package resolution.
+
+This keeps the developer-facing model close to today’s React Native workflow: dependencies are still discovered from the JavaScript project, but the integration backend is SPM instead of CocoaPods.
+Local modules
```
tibbe: Was this intended as a heading (it's currently not rendered as such)?

tibbe (commented):
Thanks for putting this together!

What's the story for migrating existing (non-framework) setups that already have a generated XCode project, etc to the new setup?

proposals/0994-swift-package-manager-support-for-react-native-ios-projects.md
chrfalch: Current version automatically updates before building throug an Xcode script.

MikePendo:
Hi 
wanted to raise another use case that we are currently struggeled with and I want you to take in to account.
When a dynamic framework is resolved via SPM inside a subproject (like `Pods.xcodeproj`), Xcode fails to automatically embed and sign it in the main app bundle. It compiles fine, but crashes instantly on physical devices with `dyld: Library not loaded`.

chrfalch:
> Hi wanted to raise another use case that we are currently struggeled with and I want you to take in to account. When a dynamic framework is resolved via SPM inside a subproject (like `Pods.xcodeproj`), Xcode fails to automatically embed and sign it in the main app bundle. It compiles fine, but crashes instantly on physical devices with `dyld: Library not loaded`.

Could you give a simple reproduction for this so that we could learn more?

MikePendo:
Hi @chrfalch — thanks for looking into this.
(sorry for late response I will genrate the message with AI so you could feed it and easily reproduce)

 Our React Native plugin's SPM support isn't released yet, so I can't point you at a published package that wires this up. But you don't need it — the crash is a pure Xcode/SPM/dyld issue (not React Native specific), and everything needed to reproduce is already public. Here's a compact spec you can hand to an AI/agent.

### Bug
A **dynamic** framework resolved via SPM inside a **subproject** (e.g. `Pods.xcodeproj`) is linked but **not** auto-embedded/signed into the app bundle. It compiles, runs on Simulator, but crashes at launch on a physical device with `dyld: Library not loaded`.

### Reproduction (no React Native, no account, no unreleased code)
Our native SDK is a public, standard SPM binary package (a `.binaryTarget` **dynamic** XCFramework), so plain Xcode is enough:

1. Create a plain iOS app target (`ReproApp`).
2. Add a second Xcode project as a **subproject** in the workspace (name it `Pods.xcodeproj` to mirror the CocoaPods layout).
3. Add this **public** SPM package, resolved on the **subproject** — NOT directly on the app target:
   - URL: `https://github.com/pendo-io/pendo-mobile-sdk.git`
   - Product: `Pendo`
4. Reference a `Pendo` symbol from the app so it links (e.g. `import Pendo`). Do **not** add it under the app target's "Frameworks, Libraries, and Embedded Content" with Embed & Sign — that auto-fixes it and hides the bug.
5. Build & run on a **physical device**.

**Expected:** app launches.
**Actual:** crash at launch → `dyld: Library not loaded: @rpath/Pendo.framework/Pendo … Reason: image not found`.

### Constraints that must hold (otherwise the bug hides)
- Framework must be **dynamic** (Pendo is).
- Package must be resolved in a **subproject**, not directly on the app target.
- Must run on a **device** — the Simulator masks it.

### Root cause
Pure CocoaPods embeds dynamic frameworks via its generated `[CP] Embed Pods Frameworks` phase (`Pods-<App>-frameworks.sh`), which copies each framework into `Frameworks/` and re-signs it. Once the dependency is SPM-resolved instead of a Pod, it's not in that script — and Xcode only auto-embeds SPM products attached **directly to the app target**, not ones resolved inside a subproject. So nothing copies/signs it into the bundle.

### RFC implication
Declaring the SPM dependency isn't sufficient on its own — there needs to be a defined embed + sign story for **dynamic** frameworks resolved in subprojects, otherwise every dynamic-framework SDK will crash on device.

Let me know if a clone-and-run repro repo would help and I'll put one together.

chrfalch:
@MikePendo - The good news is that it comes from the hybrid layout: the framework being resolved on a subproject (Pods.xcodeproj) rather than on the app target. As you note in step 4, attaching it directly to the app target with Embed & Sign "auto-fixes it and hides the bug" — and that's exactly the configuration this RFC produces. add injects the SPM package references directly into the app's existing .xcodeproj / app target, in place, and add --deintegrate runs pod deintegrate and removes React Native from the Podfile, so in the end state there's no Pods.xcodeproj subproject holding the dependency. Xcode's automatic embed-and-sign for products attached to the app target is what covers the dynamic-framework case here.

Where your point does land squarely is brownfield and the migration window, where an SPM-resolved dynamic framework can still live in a subproject alongside an existing CocoaPods/Xcode setup. I'll capture "embed + sign story for dynamic frameworks resolved in subprojects" as an explicit requirement for the brownfield/migration tooling so we don't regress into this on device. A clone-and-run repo would still be genuinely useful there — if you're up for putting one together, please do.

PR #1002, opened 2026-05-21
RFC: Privatize core-cli-utils package

huntie:
Proposal to move [`@react-native/core-cli-utils`](https://www.npmjs.com/package/@react-native/core-cli-utils) from `packages/` to `private/` and stop publishing to npm. The package has no external consumers and is better framed as a reference implementation of React Native CLI tooling than a public API.

Screenshot 2026-05-21 at 15 00 19

### DIFF
--- a/proposals/1002-privatize-core-cli-utils.md
+++ b/proposals/1002-privatize-core-cli-utils.md
@@ -1,0 +1,73 @@
+---
+title: Move core-cli-utils to private/ as a reference implementation
+author:
+  - Alex Hunt
+date: 2026-05-21
+---
+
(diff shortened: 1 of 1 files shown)


cortinico:
I've done a pass here:
https://github.com/search?type=code&q=%22%40react-native%2Fcore-cli-utils%22+NOT+is%3Afork+NOT+org%3Afacebook+%28language%3AJavaScript+OR+language%3AJSON+OR+language%3ATypeScript%29&p=2

And it seems there are no real usages in OSS of this package so I think we should be safe moving it to private ✅

cortinico (approved):

PR #1003, opened 2026-05-23
RFC: Add dedicated Instance types for built-in component refs (JS API stabilization)

huntie:
Proposes introducing dedicated `*Instance` type exports (e.g. `ViewInstance`, `TextInputInstance`) for every built-in React Native component, providing a canonical way to type refs.

**Why this matters**: This fills a gap in the public API that was previously hidden by class declarations in the hand-written TypeScript types, and provides a concise migration path (`useRef<View>` → `useRef<ViewInstance>`) in our from-source Strict TypeScript API — currently the **highest-friction breaking change** for users.

# [View the rendered RFC](https://github.com/react-native-community/discussions-and-proposals/blob/main/proposals/1003-strict-api-component-instance-types.md)

### DIFF
--- a/proposals/1003-strict-api-component-instance-types.md
+++ b/proposals/1003-strict-api-component-instance-types.md
@@ -1,0 +1,308 @@
+---
+title: Add dedicated Instance types for built-in component refs (JS API)
+author:
+  - Alex Hunt
+date: 2026-05-23
+---
+
+# RFC1003: Add dedicated Instance types for built-in component refs (JS API)
+
+## Summary
+
+Add a publicly exported `*Instance` type for every built-in React Native component (e.g. `ViewInstance`, `TextInputInstance`, `ScrollViewInstance`), providing a canonical way to type refs. This is a missing piece of our API, historically hidden by the use of class declarations in the hand-written TypeScript types.
+
+## Basic example
+
+```tsx
+import { useRef } from "react";
+import type { ViewInstance, TextInputInstance } from "react-native";
+
+function MyComponent() {
+  const viewRef = useRef<ViewInstance>(null);
+  const inputRef = useRef<TextInputInstance>(null);
+
+  return (
+    <>
+      <View ref={viewRef} />
+      <TextInput ref={inputRef} />
+    </>
+  );
+}
+```
+
+These types are also valid for `Animated` variants — the `Animated` wrapper is transparent for refs:
+
+```tsx
+const viewRef = useRef<ViewInstance>(null);
+
+<Animated.View ref={viewRef} />;
+```
+
+## Motivation
+
+Typing refs for built-in components is harder than it should be, and it's about to get harder.
+
+- **No canonical pattern**: There is currently no canonical way to type a ref to a built-in React Native component.
+- **Three competing approaches**: Developers face three different patterns depending on the language, the types version, and which documentation they read. None is satisfying on its own.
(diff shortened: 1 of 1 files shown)


proposals/1003-strict-api-component-instance-types.md
```diff
@@ -0,0 +1,308 @@
...
+}
+```
+
+**💡 Key fix**: These specialised instance types are the cases where `HostInstance` silently loses type safety today. Dedicated `*Instance` types ensure developers get autocomplete and type checking for component-specific imperative methods.
+
+### Naming convention
+
+- Pattern: `{ComponentName}Instance` (e.g. `ViewInstance`, `TextInputInstance`).
+- This aligns with React Native's existing naming (`HostInstance`, `TextInputInstance` which already exists internally) and the `Instance` terminology used in React's own types (`React.ComponentRef`, `React.ElementRef`).
+- We intentionally diverge from the web convention (`HTMLDivElement`, `HTMLInputElement`) because React Native elements are not HTML elements. The `*Instance` suffix also avoids ambiguity with the `ReactNativeElement` base class.
```
rubennorte: I'm not sure this is accurate. We kinda have HTML elements (technically we implement a subset of the DOM interface).

Also not sure why we should avoid ambiguity with the `ReactNativeElement` base class, given these are all instances of that or a subclass (so calling them something like `RNImageElement` or just `ImageElement` would have good symmetry with what's done on Web).

proposals/1003-strict-api-component-instance-types.md
```diff
@@ -0,0 +1,308 @@
...
+  getScrollResponder(): ScrollResponderType;
+  // ... etc.
+}
+```
+
+**💡 Key fix**: These specialised instance types are the cases where `HostInstance` silently loses type safety today. Dedicated `*Instance` types ensure developers get autocomplete and type checking for component-specific imperative methods.
+
+### Naming convention
+
+- Pattern: `{ComponentName}Instance` (e.g. `ViewInstance`, `TextInputInstance`).
```
rubennorte: I think we should consider the alternative "Element" suffix, either as `ViewElement` or as `RNViewElement`. Doing it now might be better than doing it later, as we already now that the "Instance" terminology isn't used on Web and it's kinda legacy here.

rubennorte (approved):
Thanks for writing this down, @huntie ! The proposal is very well articulated and I think the change is very safe. I'd only reconsider the naming, aligning with Web now that we already have DOM APIs in RN.

proposals/1003-strict-api-component-instance-types.md
huntie: Yeah, rephrasing this ("avoids ambiguity..." bit is definitely wrong 👍🏻) — it's mostly about the currently exported `HostInstance` type and its TS/Flow usages.

- It would instead _**remove**_ a layer of indirection to remove `*Instance` as `HostInstance = ReactNativeElement` currently.
- The remaining issue is the naming churn now vs eventual (`*Element` types will live alongside `HostInstance` and others). But yeah perhaps we should optimise for the most correct naming regardless.

huntie: The context of RN having more DOM-like APIs on these types is useful too 👍🏻

huntie:
**🔁 Back to `*Instance` again**

Caught up offline with @rubennorte and we noted two extra advantages of `*Instance` that tilt the scales:

- "Instance" is conceptually one layer derived from `ReactNativeElement`, and these are not necessarily 1:1 with underlying native elements (e.g. `VirtualizedList` cells).
- Avoid churn (avoid `HostInstance` deprecation + exposing `ReactNativeElement` + potential internal flattening) (also symmetry with `HostComponent`). As above, `Instance` types give us a nice layer of abstraction on the public instance types.

PR #1006, opened 2026-06-09
feat: agp v9 adoption

hurali97:
Proposal: Android Gradle Plugin v9 Adoption

### DIFF
--- a/proposals/0000-android-agp-v9-adoption.md
+++ b/proposals/0000-android-agp-v9-adoption.md
@@ -72,7 +72,7 @@ To measure AGP v9 impact, we ran a discovery phase by enabling AGP v9 on React N
 
 For React Native core, the following changes were required in `react-native-community/template` and `facebook/react-native`:
 
-**react-native-community/template**:
+### react-native-community/template
 
 - Upgrade to Gradle v9.4.1
   - This is already addressed in https://github.com/react-native-community/template/pull/227
@@ -83,7 +83,7 @@ For React Native core, the following changes were required in `react-native-comm
 - Enable opt outs in `gradle.properties`
   - This is already addressed in https://github.com/react-native-community/template/pull/224
 
-**facebook/react-native**:
+### facebook/react-native
 
 - Upgrade to Gradle v9.4.1
   - This is already addressed in https://github.com/facebook/react-native/pull/57078
@@ -98,12 +98,14 @@ For React Native core, the following changes were required in `react-native-comm
 
 This completes Phase 1 described above.
 
+<hr/>
+
 The second area is AGP v9 compatibility for community libraries. Below are the common migration patterns and library-specific notes.
 
-**Community libraries**:
+### Community libraries
 
 Below is the matrix used during discovery:
-
+  
 | Library | Version | Patch link |
 | --- | --- | --- |
 | react-native-reanimated | v4.3.1 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/react-native-reanimated%2B4.4.1.patch |
@@ -121,6 +123,8 @@ Below is the matrix used during discovery:
 | async-storage | v3.1.0 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/%40react-native-async-storage%2Basync-storage%2B3.1.0.patch |
 | nitro-image | v0.14.0 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/react-native-nitro-image%2B0.14.0.patch |
 
+<hr/>
+
 - Most libraries apply the `kotlin-android` plugin in `build.gradle`. With built-in Kotlin, this is no longer needed and can fail builds.
 - Previously, `java.srcDirs` plus `kotlin-android` covered Kotlin-only sources. With built-in Kotlin, Kotlin-only sources should be defined via `kotlin.srcDirs`.
 - `.srcDirs` is deprecated in favor of `.directories`.
@@ -131,8 +135,14 @@ The points above cover common migration work. One notable library-specific case:
 
 - The `nitrogen` template (used by Nitro modules) defines `java.srcDirs` for Kotlin-only sources. With built-in Kotlin this must move to `kotlin.directories.add`.
 
+</details>
+
+<hr/>
+
 **Expo**:
 [TBA]
+
+<hr/>
 
 ## Adoption strategy
 
--- a/proposals/0000-android-agp-v9-adoption.md
+++ b/proposals/0000-android-agp-v9-adoption.md
@@ -128,7 +128,7 @@ Below is the matrix used during discovery:
 
 <hr/>
 
-- Most libraries apply the `kotlin-android` plugin in `build.gradle`. With built-in Kotlin, this is no longer needed and can fail builds.
+- Most libraries apply the `kotlin-android` (or `org.jetbrains.kotlin.android`) plugin unconditionally in `build.gradle`. AGP 9 registers the `kotlin` extension itself, so this now fails. The backward-compatible fix is to apply the plugin only when AGP has not already provided it: `if (project.extensions.findByName('kotlin') == null) { apply plugin: 'kotlin-android' }`, placed after `apply plugin: 'com.android.library'`.
 - Previously, `java.srcDirs` plus `kotlin-android` covered Kotlin-only sources. With built-in Kotlin, Kotlin-only sources should be defined via `kotlin.srcDirs`.
 - `.srcDirs` is deprecated in favor of `.directories`.
 - `android.kotlinOptions` is no longer valid with built-in Kotlin; use `kotlin.compilerOptions`.
--- a/proposals/0000-android-agp-v9-adoption.md
+++ b/proposals/0000-android-agp-v9-adoption.md
@@ -31,6 +31,9 @@ This RFC proposes a three-phase AGP v9 adoption strategy:
 - Roll out AGP v9 compatibility changes to actively used libraries while maintaining backward compatibility.
 - Coordinate with the community for broader ecosystem coverage.
 - Align with Expo's AGP v9 adoption efforts in Expo core and related libraries.
+- Create a new library plugin from RNGP while ensuring backward compat.
+  - The RNGP will be restricted to be used only by the App's.
+  - This is to allow distributing and maintaining library specific changes independent of RNGP.
 
 #### Phase 3
 - Remove the AGP v9 opt-outs from the React Native template.
@@ -144,17 +147,39 @@ The points above cover common migration work. One notable library-specific case:
 
 <hr/>
 
+### Separate Library Plugin
+
+Currently, the react-native applications and libraries are both handled by the RNGP(`com.facebook.react`) to apply auto-linking, codegen changes, etc. This means that the app and library specific handling is fused within the RNGP and doing a major change only for library means we ensure the app handling is not breaking, so to avoid this friction, we now create a new library gradle plugin by extracting pieces from RNGP and putting it together as `com.facebook.react.library`.
+
+**Details**
+- Introduce a new module within `gradle-plugins` and register it, so it can be used via comosite build
+  - Since it will be used via composite build, we do not expect any change in user's `settings.gradle`
+- Extract the library pieces out to a separate gradle plugin
+  - This means if there are any code in RNGP that the library plugin needs, we duplicate that implementation.
+  - This is a temporary trade-off until we fully deprecate RNGP for librares, see below.
+- Add deprecation notice to RNGP to warn the users over a new couple releases to start using dedicated `com.facebook.react.library` plugin.
+  - Upon reaching the deprecation window, we remove the library specific code from RNGP.
+
+> A draft version of this can be found here - https://github.com/react/react-native/pull/57912
+
+**Community**
+- This will require us to update `react-native-builder-bob` to use `com.facebook.react.library` by default.
+
+<hr/>
+
 ## Adoption strategy
 
 <!-- If we implement this proposal, how will existing React Native developers adopt it? Is this a breaking change? Can we write a codemod? Should we coordinate with other projects or libraries? -->
 
-As described above, AGP v9 adoption is rolled out in three phases. Phase 1 is complete and should ship with React Native v0.87.x. Newly created apps do not need changes because template opt-outs are enabled by default. Existing apps upgrading to v0.87.x should apply a few manual updates described in the next section. At this stage, we expect no breakages with actively maintained community libraries.
+As described in the earlier section, AGP v9 adoption is rolled out in three phases. Phase 1 is complete and should ship with React Native v0.87.x. Newly created apps do not need changes because template opt-outs are enabled by default. Existing apps upgrading to v0.87.x should apply a few manual updates described in the next section. At this stage, we expect no breakages with actively maintained community libraries.
 
 In Phase 2, we will submit compatibility pull requests to actively used and maintained libraries from the matrix above while preserving backward compatibility. Broader ecosystem coverage will require community participation.
 
 As part of Phase 2, the Expo team will evaluate required changes on their side, including ecosystem impact.
 
-Phase 2 needs sufficient time, especially for community adoption. One possible support mechanism is adding an "AGP v9 compatible" filter in `react-native-community/directory` to track progress.
+The separate library gradle plugin will also be landing as part of the Phase 2.
+
+Phase 2 needs sufficient time, especially for community adoption.
 
 Once there is enough confidence and ecosystem coverage, we can move to Phase 3: remove opt-outs from `react-native-community/template` and ask existing apps to remove them as well. At that point, the ecosystem should fully leverage built-in Kotlin, new DSL behavior, and other AGP v9 defaults.
 
@@ -183,7 +208,16 @@ android.newDsl=false
 
 In addition to manual instructions, we should update `react-native-community/upgrade-helper` so this guidance is included by default.
 
-In Phase 2, we will migrate widely used libraries. To accelerate this, we should include a public request for help in the v0.87.x release post and X announcements.
+In Phase 2, we will migrate widely used libraries. To accelerate this, we should include a public request for help in the v0.x.x release post and X announcements.
+
+As part of the release post, we should add section about deprecating library support from RNGP and recommend using `com.facebook.react.library`. The section should include:
+
+- This is a breaking change for libraries.
+- If this change is shipped, say in 0.89.0 release, the apps when they upgrade to this version will see a deprecation notice for each third party library not using `com.facebook.react.library`. This allows the community a decent window to start adopting the new library gradle plugin.
+- In the next few minor releases, the RNGP support for libraries will be removed completely.
+- We will backport the change to the react-native versions in the release window.
+- The react-native versions outside of the release window, will not work with the new library plugin.
+  - The third party libraries may need to allow their users to use a version of their library which is compatible with the react-native version below the lower bound of release window. (For eg, 0.85.0 is the lower bound for 0.87.0 main release)
 
 For Phase 3, we should announce opt-out removal from `react-native-community/template` in the v0.88.x release post and state the expected removal timeline (for example, v0.89.x).
 
@@ -192,3 +226,4 @@ For Phase 3, we should announce opt-out removal from `react-native-community/tem
 <!-- Optional, but suggested for first drafts. What parts of the design are still TBD? -->
 
 - Expo needs to evaluate its scope of AGP v9 changes and provide feedback on adoption timelines and risks.
+- When the RNGP support for libraries will be removed completely? (Ideally, a time frame.)
--- a/proposals/0000-android-agp-v9-adoption.md
+++ b/proposals/0000-android-agp-v9-adoption.md
@@ -1,0 +1,184 @@
+---
+title: Android - AGP v9 adoption
+author:
+- Hur Ali
+date: 2026-06-09
+---
+
+# RFC0000: Android - AGP v9 adoption
+
+## Summary
+
+<!-- Brief explanation of the change. -->
+
+Android Gradle Plugin [v9.0.1](https://developer.android.com/build/releases/agp-9-0-0-release-notes) became stable in January 2026. AGP v9 introduces changes that affect React Native core, community libraries, and consumer applications. AGP v9 provides temporary opt-outs in `gradle.properties`, but those opt-outs will be removed in [AGP v10](https://developer.android.com/build/releases/gradle-plugin-roadmap#agp-10), expected in late 2026.
+
+Key breaking changes include:
+- Built-in Kotlin support in AGP v9. Projects should not explicitly apply `"org.jetbrains.kotlin.android"` once this mode is enabled. This is high impact because many libraries and apps currently apply it.
+- With built-in Kotlin, `java.srcDirs` cannot be used for Kotlin-only sources. Also, `srcDirs` is deprecated in favor of `.directories.add`.
+- `android.kotlinOptions` is removed in favor of `kotlin.compilerOptions`.
+- `android.applicationVariants` and `android.libraryVariants` are removed.
+- `com.android.build.gradle.LibraryExtension` is deprecated.
+
+This RFC proposes a three-phase AGP v9 adoption strategy:
+
+#### Phase 1
+- Enable opt-outs and use the optimized ProGuard file in `react-native-community/template`.
+- Upgrade to AGP v9.2.0 and Gradle v9.4.1 in both the template and React Native core.
+- Ensure React Native core can build and run with AGP v9.
+
+#### Phase 2
+- Roll out AGP v9 compatibility changes to actively used libraries while maintaining backward compatibility.
+- Coordinate with the community for broader ecosystem coverage.
+- Align with Expo's AGP v9 adoption efforts in Expo core and related libraries.
+
+#### Phase 3
+- Remove the AGP v9 opt-outs from the React Native template.
+
+## Basic example
+
+<!-- If the proposal involves a new or changed API, include a basic code example. Omit this section if it's not applicable. -->
+
+In Phase 1, application `gradle.properties` should include two temporary opt-out flags:
+
+```gradle.properties
+# Opt out of built-in kotlin and new DSL behavior that ships with AGP 9.
+# Starting from AGP 10.x these opt outs will be removed.
+android.builtInKotlin=false
+android.newDsl=false
+```
+
+In Phase 3, these properties will be removed from `react-native-community/template`.
+
+## Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected outcome?
+
+Please focus on explaining the motivation so that if this RFC is not accepted, the motivation could be used to develop alternative solutions. In other words, enumerate the constraints you are trying to solve without coupling them too closely to the solution you have in mind. -->
+
+Keeping React Native aligned with current AGP versions is required for long-term ecosystem health. Without this upgrade, we risk:
+
+- Consumers being unable to compile Android apps with newer SDK targets, and eventually unable to ship updates when Google Play requires newer target SDK levels. Each AGP version has an upper bound on supported compile/target SDK versions.
+
+- Android Studio build incompatibility. Projects using AGP versions outside the [supported compatibility window](https://developer.android.com/build/releases/about-agp#android_gradle_plugin_and_android_studio_compatibility) may fail to build in newer IDE versions.
+
+- Growing divergence with community Android libraries as they move to AGP v9+, increasing maintenance burden and risking delayed adoption of security fixes.
+
+## Detailed design
+
+<!-- This is the bulk of the RFC. Explain the design in enough detail for somebody familiar with React Native to understand, and for somebody familiar with the implementation to implement. This should get into specifics and corner-cases, and include examples of how the feature is used. Any new terminology should be defined here. -->
+
+To measure AGP v9 impact, we ran a discovery phase by enabling AGP v9 on React Native v0.85.3 and validating a set of community libraries. This produced a concrete list of required changes across React Native core and ecosystem packages.
+
+For React Native core, the following changes were required in `react-native-community/template` and `facebook/react-native`:
+
+**react-native-community/template**:
+
+- Upgrade to Gradle v9.4.1
+  - This is already addressed in https://github.com/react-native-community/template/pull/227
+
+- Use optimized ProGuard file
+  - This is already addressed in https://github.com/react-native-community/template/pull/223
+
+- Enable opt outs in `gradle.properties`
+  - This is already addressed in https://github.com/react-native-community/template/pull/224
+
+**facebook/react-native**:
+
+- Upgrade to Gradle v9.4.1
+  - This is already addressed in https://github.com/facebook/react-native/pull/57078
+
+- Adopt AGP v9
+  - Upgrade to AGP v9.2.0
+  - Remove `alias(libs.plugins.kotlin.android)` in favor of built-in Kotlin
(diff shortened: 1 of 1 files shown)


proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,184 @@
...
+- `.srcDirs` is deprecated in favor of `.directories`.
+- `android.kotlinOptions` is no longer valid with built-in Kotlin; use `kotlin.compilerOptions`.
+- `fix-prefab.gradle` scripts in C++ (CMake) libraries often use removed APIs (`android.applicationVariants` / `android.libraryVariants`); migrate to `androidComponents.onVariants`.
+
+The points above cover common migration work. One notable library-specific case:
+
+- The `nitrogen` template (used by Nitro modules) defines `java.srcDirs` for Kotlin-only sources. With built-in Kotlin this must move to `kotlin.directories.add`.
+
+**Expo**:
+[TBA]
```
Kudo, line 135: thanks for having the great rfc. i did have some attemts for expo migration. maybe we could try to pull the change with backward compatiblity in mind at our end first.

- https://github.com/Kudo/sdk56agp9/blob/main/patches/expo-modules-autolinking%4056.0.13.patch
- https://github.com/Kudo/sdk56agp9/blob/main/patches/expo-modules-core%4056.0.13.patch

hurali97: Thanks for sharing these. Perhaps you can sync with @lukmcall - I asked him about Expo state before this RFC. 

I have already forwarded this RFC and mentioned your comments to him on Discord.

lukmccall:
The RFC looks great, thanks for putting it together! I've already started working on AGP9 support on the Expo side and opened a few PRs. Most things are working well, but I've hit a few unexpected issues that'll need more work. For instance, we have some custom Apollo client configuration in the dev-launcher that stopped working. I'll share more once I figure out how to fix it.

lukmccall:
One random thought: what do you guys think about creating a `kotlin-android-compat` Gradle plugin that would apply `kotlin-android` conditionally based on the AGP version and whether the built-in Kotlin is enabled? It would make migration a bit easier for library authors, as most packages could then be compatible with both AGP 8 and 9.

proposals/0000-android-agp-v9-adoption.md
Kudo: all good! he has more insight and expertise than me. i'm glad that you have talked already

hurali97:
> One random thought: what do you guys think about creating a `kotlin-android-compat` Gradle plugin that would apply `kotlin-android` conditionally based on the AGP version and whether the built-in Kotlin is enabled? It would make migration a bit easier for library authors, as most packages could then be compatible with both AGP 8 and 9.

This sounds like a nice idea and could help with one part of the migration, but after thinking about it more, I’m not sure it would add enough value on its own.

Even with a kotlin-android-compat plugin, library authors would still need to handle a few manual migration steps:

- Remove kotlin-android / org.jetbrains.kotlin.android from their build.gradle for AGP 9+, since built-in Kotlin replaces that plugin there.
- Remove or migrate deprecated/removed APIs such as android.kotlinOptions and usages of .srcDirs where they are no longer valid.
- Avoid configuring Kotlin-only sources through java.srcDirs.
- Add AGP-version checks for the remaining compatibility differences anyway.

The custom plugin could dynamically apply kotlin-android for AGP < 9, which is useful. But it cannot really solve the case where the consuming project has already applied kotlin-android, because we can apply plugins programmatically, but we can’t safely “unapply” a plugin that was already applied by the build script.

So in practice, the plugin would cover only one part of the migration: conditionally applying kotlin-android for AGP 8 and below. The removal of kotlin-android for AGP 9+, plus the other DSL/source-set migrations, would still require manual changes from library authors. Because of that, I think the value might be relatively limited.

lukmccall:
Yes, I agree it won't solve every problem - realistically, this will still require some manual changes from library maintainers. But other changes, such as using `kotlin.srcDirs` instead of `java.srcDirs`, should work with both AGP versions. I'm only suggesting we add it to make supporting both possible in the majority of cases with minimal effort from the maintainer's perspective. The issue is that if you're using the plugins block, you can't apply plugins conditionally without falling back to the old apply plugin syntax - such a plugin would solve that.

hurali97:
>The issue is that if you're using the plugins block, you can't apply plugins conditionally without falling back to the old apply plugin syntax 

True that. So If I catch it correctly, the library maintainers will need to replace the `kotlin-android` plugin with compat Gradle plugin? 

For some libraries, this might be the only change that they need, for eg, `react-native-pager-view`. However, for majority the authors will need other manual adjustments.

I am also wondering whether to introduce a separate compat plugin OR we can add this as a fallback mechanism to RNGP? That way the library authors may not need to implement another plugin.

Nicola [suggested](https://app.notion.com/p/AGP9-Discovery-and-Adoption-3675d027c0f8801da10ccab06b4e9322?d=0065d027c0f883499535830ce0a70778&source=copy_link) that we can also think of separating library implementations from RNGP in a separate Gradle plugin, however that is optional for now and we will start that separately than AGP 9. I am mentioning it here because if we are able to add this compat mechanism to RNGP, we can later extract it out.

>  But other changes, such as using kotlin.srcDirs instead of java.srcDirs, should work with both AGP versions

Yes it should work, however there are a few catches. The kotlin-only sources can not be added to `java.srcDirs`, they should be added to `kotlin.srcDirs`, otherwise with built-in kotlin on AGP 9 there is a build crash. The `.srcDirs` is deprecated in favor of `.directories` with AGP v9, so in the future versions this maybe removed.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Adoption strategy
+
+<!-- If we implement this proposal, how will existing React Native developers adopt it? Is this a breaking change? Can we write a codemod? Should we coordinate with other projects or libraries? -->
+
+As described above, AGP v9 adoption is rolled out in three phases. Phase 1 is complete and should ship with React Native v0.87.x. Newly created apps do not need changes because template opt-outs are enabled by default. Existing apps upgrading to v0.87.x should apply a few manual updates described in the next section. At this stage, we expect no breakages with actively maintained community libraries.
+
+In Phase 2, we will submit compatibility pull requests to actively used and maintained libraries from the matrix above while preserving backward compatibility. Broader ecosystem coverage will require community participation.
+
+As part of Phase 2, the Expo team will evaluate required changes on their side, including ecosystem impact.
```
cortinico, line 155: Was there an update on this front? @lukmccall @Kudo ?

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,194 @@
...
+
+<hr/>
+
+## Adoption strategy
+
+<!-- If we implement this proposal, how will existing React Native developers adopt it? Is this a breaking change? Can we write a codemod? Should we coordinate with other projects or libraries? -->
+
+As described above, AGP v9 adoption is rolled out in three phases. Phase 1 is complete and should ship with React Native v0.87.x. Newly created apps do not need changes because template opt-outs are enabled by default. Existing apps upgrading to v0.87.x should apply a few manual updates described in the next section. At this stage, we expect no breakages with actively maintained community libraries.
+
+In Phase 2, we will submit compatibility pull requests to actively used and maintained libraries from the matrix above while preserving backward compatibility. Broader ecosystem coverage will require community participation.
```
cortinico, line 153: >  while preserving backward compatibility

This is the key of the whole proposal. I'm wondering how can we update libraries to don't apply `kotlin-android` while still retaining backward compat.

In the past we talked about a `com.facebook.react.library` Gradle Plugin that is versioned **outside** the react/react-native repo. Is this something we're still interested in driving?

If so let's add it in this RFC. cc @lukmccall @Kudo

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,194 @@
...
+
+As part of Phase 2, the Expo team will evaluate required changes on their side, including ecosystem impact.
+
+Phase 2 needs sufficient time, especially for community adoption. One possible support mechanism is adding an "AGP v9 compatible" filter in `react-native-community/directory` to track progress.
```
cortinico, line 157: If possible I would avoid the extra filter in the directory (that should be doable if we implement this in being fully backward compat).

cortinico (approved):
Great RFC @hurali97 (sorry for the late review).

I've left some comments. I know that @lukmccall @Kudo have some updates on the Expo front to share also

proposals/0000-android-agp-v9-adoption.md
lukmccall: On our side, we're ready. All Expo code should work with AGP 9 and be compatible with build-in Kotlin. There might be some leftovers that I missed. I'll retest everything whenever RN officially starts supporting AGP 9.

proposals/0000-android-agp-v9-adoption.md
lukmccall: We can apply kotlin-android conditionally using our own Gradle plugin - it would look like this: https://github.com/expo/expo/blob/7d1359bd667168ca9ae64afc17058e558df23163/packages/expo-modules-core/expo-module-gradle-plugin/src/main/kotlin/expo/modules/plugin/ProjectConfiguration.kt#L23-L33. Not sure if we should combine this RFC with the RN turbomodules Gradle plugin. Maybe it would be better to publish `kotlin-android-compat` first to only address the AGP migration, as proposed above: https://github.com/react-native-community/discussions-and-proposals/pull/1006#issuecomment-4680920576

hurali97: On our end, as part of a POC, we've separated the Gradle setup into `RNGP` for app consumption and `com.facebook.react.library` for library consumption. This separation currently lives in RN Core, but it can be extracted. I need a bit of time to finalize the implementation and open a draft PR, after which we can align on the exact structure.

A few key points on this approach:

Backward Compatibility: I’ve kept this backward compatible for now, meaning libraries can still apply `RNGP`. Though whether we maintain this fallback is open for debate.

Plugin Integration: A clear benefit is that we can directly apply the `kotlin-android-compat` plugin inside `com.facebook.react.library` for libraries to use.

One caveat for AGP 9 compatibility: even with the compat plugin, library maintainers will still need to manually remove the `kotlin-android` plugin.

proposals/0000-android-agp-v9-adoption.md
cortinico: > On our side, we're ready

How is expo going to handle the removal of the opt-out?

lukmccall: From the module's perspective, the Expo Gradle plugin handles all 3 cases: AGP 8, AGP 9 with built-in Kotlin opted out, and AGP 9 with built-in Kotlin support.

For users who use CNG, we'll apply the changes in the app template automatically. For the rest, they'll have to adjust their apps themselves (we can't do anything more here).

cortinico: @lukmccall My question was more related to the 3p libraries that have imports of `kotlin-android` which will break the build (once we remove the opt-out)

lukmccall: People will have to migrate. We can't automate that process in any way. However, we can make it easier - that's why I proposed the `kotlin-android-compat` plugin as a temporary workaround to support both cases (with and without opt-out). But yes, that requires 3rd party libraries to migrate. We can help in that process by making PRs. If someone is using Expo Modules, they'll work as long as the maintainers have updated the template, which is a year old.

cortinico: Yeah exactly. Can we expand the RFC with the `kotlin-android-compat` plugin and how we're planning to distribute it? cc @hurali97

hurali97: We can expand the RFC, but we’ll need details on how Expo plans to publish and handle the usage of this plugin. cc: @lukmccall 

Additionally, I think this plugin should remain optional, especially for non Expo projects. If library maintainers have already added backward compatibility for AGP v9 or simply prefer to manage it manually, they should be free to do so.

We should also highlight this in the upcoming React Native blog post section on AGP v9 adoption: developers can either use the compatibility plugin or handle backward compatibility themselves.

lukmccall: > We can expand the RFC, but we’ll need details on how Expo plans to publish and handle the usage of this plugin

Hmm, I think we should publish that plugin to Maven under the React Foundation - similar to how React Native is published. Publishing it under Expo seems weird, as in our case we don't need it at all. However, if you think that would be easier, we can handle it on our side.

> Additionally, I think this plugin should remain optional, especially for non Expo projects. If library maintainers have already added backward compatibility for AGP v9 or simply prefer to manage it manually, they should be free to do so.

I agree that the plugin should remain optional

hurali97: > Hmm, I think we should publish that plugin to Maven under the React Foundation - similar to how React Native is published. Publishing it under Expo seems weird, as in our case we don't need it at all. However, if you think that would be easier, we can handle it on our side.

If we're planning to land a separate library plugin anyway, we could bundle the kotlin-android-compat logic directly into it rather than publishing a separate plugin. Since tomorrow's draft PR for separate library plugin will likely need a few iterations. Publishing the standalone kotlin-android-compat plugin first might be the faster route. However, if we're flexible with the timeline, encapsulating that logic inside library plugin and have only this applied by the consumers. cc: @cortinico

cortinico: > If we're planning to land a separate library plugin anyway, we could bundle the kotlin-android-compat logic directly into it rather than publishing a separate plugin

Yes ideally that's the preferred outcome.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+
+- Android Studio build incompatibility. Projects using AGP versions outside the [supported compatibility window](https://developer.android.com/build/releases/about-agp#android_gradle_plugin_and_android_studio_compatibility) may fail to build in newer IDE versions.
+
+- Growing divergence with community Android libraries as they move to AGP v9+, increasing maintenance burden and risking delayed adoption of security fixes.
+
+## Detailed design
+
+<!-- This is the bulk of the RFC. Explain the design in enough detail for somebody familiar with React Native to understand, and for somebody familiar with the implementation to implement. This should get into specifics and corner-cases, and include examples of how the feature is used. Any new terminology should be defined here. -->
+
+To measure AGP v9 impact, we ran a discovery phase by enabling AGP v9 on React Native v0.85.3 and validating a set of community libraries. This produced a concrete list of required changes across React Native core and ecosystem packages.
```
gabrieldonadel: [💡 suggestion] We ran a larger discovery pass that can back this section with data. The harness enabled both AGP 9 defaults (`android.newDsl=true`, `android.builtInKotlin=true`) and built 1002 community packages, one Gradle build each.

Results: 364 `pass`, 281 `fail-newdsl` (fails with the flags on, builds with them off), 348 `fail-baseline`.

Note one caveat. 72 of the passes are Capacitor or Cordova plugins. They produce no Gradle tasks, so React Native never builds them. The meaningful pass count is 292.

This dataset can turn the discovery section from a 14-library sample into a usage-weighted view of the whole ecosystem. We can share the full per-package results and logs.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+| react-native-nitro-modules | v0.35.7 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/react-native-nitro-modules%2B0.35.7.patch |
+| react-native-svg | no issues - v15.15.5 - Pure java | N/A |
+| react-native-mmkv | v4.3.1 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/react-native-mmkv%2B4.3.1.patch |
+| react-native-share | no issues - v12.3.1 | N/A |
+| async-storage | v3.1.0 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/%40react-native-async-storage%2Basync-storage%2B3.1.0.patch |
+| nitro-image | v0.14.0 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/react-native-nitro-image%2B0.14.0.patch |
+
+<hr/>
+
+- Most libraries apply the `kotlin-android` plugin in `build.gradle`. With built-in Kotlin, this is no longer needed and can fail builds.
```
gabrieldonadel: [🏗️ **design**] Our data shows this one issue dominates. It is 96% of `fail-newdsl` causes (269 of 281). Another 176 cases hide inside `fail-baseline`: the flagged build died on the same collision, but the fallback build also failed for a separate reason. Counting those, the collision total is 445 across the 1002-package run.

AGP 9 registers the `kotlin` extension itself. A library that applies `kotlin-android` unconditionally then fails during configuration:

```
Cannot add extension with name 'kotlin', as there is an extension already registered with that name.
```

A backward-compatible guard fixes it without dropping support for older AGP:

```groovy
if (project.extensions.findByName('kotlin') == null) {
    apply plugin: 'kotlin-android'
}
```

The apply must come after `apply plugin: 'com.android.library'`, so AGP has already registered its extension. This form needs no version table. It also covers AGP 10, where the opt-out is removed. The RevenueCat maintainer suggested this form. We verified both flag states build clean against AGP 9.2.1.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+
+This RFC proposes a three-phase AGP v9 adoption strategy:
+
+#### Phase 1
+- Enable opt-outs and use the optimized ProGuard file in `react-native-community/template`.
+- Upgrade to AGP v9.2.0 and Gradle v9.4.1 in both the template and React Native core.
+- Ensure React Native core can build and run with AGP v9.
+
+#### Phase 2
+- Roll out AGP v9 compatibility changes to actively used libraries while maintaining backward compatibility.
```
gabrieldonadel: [💡 suggestion] Much of this Phase 2 work is already in progress and can be linked here. From the harness we filed 287 pull requests covering 311 packages. Every PR uses the backward-compatible guard shown in the community section. Current tracker state: 233 open, 50 merged, 4 closed.

Monorepos let one PR fix many packages. For example, `oblador/react-native-vector-icons` covers 13 packages in one PR.

One practical ceiling: about 82 collision packages have no reachable GitHub repository. Almost all sit at usage 0.005 or below. We can share the tracker so the RFC references concrete PRs instead of a plan.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+The points above cover common migration work. One notable library-specific case:
+
+- The `nitrogen` template (used by Nitro modules) defines `java.srcDirs` for Kotlin-only sources. With built-in Kotlin this must move to `kotlin.directories.add`.
+
+</details>
+
+<hr/>
+
+**Expo**:
+[TBA]
```
gabrieldonadel: [🏗️ **design**] We can fill this section. Two concrete Expo findings from the harness:

1. The Expo SDK 58 template already opts out of both flags in `android/gradle.properties`. This matches the Phase 1 approach in this RFC.
2. `expo-updates` (usage 0.387) does not fail on the Kotlin collision. It fails inside `expo-autolinking` on a null `KotlinJvmAndroidCompilation.getAndroidVariant()`. This is an Expo-side fix, not a per-library fix.

The Expo team is running its own AGP 9 adoption. We can add timelines and the full scope here.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+- Extract the library pieces out to a separate gradle plugin
+  - This means if there are any code in RNGP that the library plugin needs, we duplicate that implementation.
+  - This is a temporary trade-off until we fully deprecate RNGP for librares, see below.
+- Add deprecation notice to RNGP to warn the users over a new couple releases to start using dedicated `com.facebook.react.library` plugin.
+  - Upon reaching the deprecation window, we remove the library specific code from RNGP.
+
+> A draft version of this can be found here - https://github.com/react/react-native/pull/57912
+
+**Community**
+- This will require us to update `react-native-builder-bob` to use `com.facebook.react.library` by default.
```
gabrieldonadel: [💡 suggestion] There is a faster win here that does not depend on the new library plugin. Add the extension guard to the `react-native-builder-bob` default template now. That fixes the collision (the 96% case) for every newly generated library.

The guard is backward compatible. It works on current AGP 8 and older React Native. Libraries get the fix immediately, without waiting for the `com.facebook.react.library` migration window. The library-plugin migration can then proceed in parallel.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+
+<hr/>
+
+## Adoption strategy
+
+<!-- If we implement this proposal, how will existing React Native developers adopt it? Is this a breaking change? Can we write a codemod? Should we coordinate with other projects or libraries? -->
+
+As described in the earlier section, AGP v9 adoption is rolled out in three phases. Phase 1 is complete and should ship with React Native v0.87.x. Newly created apps do not need changes because template opt-outs are enabled by default. Existing apps upgrading to v0.87.x should apply a few manual updates described in the next section. At this stage, we expect no breakages with actively maintained community libraries.
```
gabrieldonadel: [🏗️ **design**] Our data suggests softening this claim. The guard removes the collision, but a second class of incompatibility appears behind it.

We rebuilt the published fixes. Of 13 fixes released to npm, only 8 build clean with the flags on. The collision (`Cannot add extension with name 'kotlin'`) is gone in all 13 logs. The 5 that still fail now fail later in the build. For example:
- `kotlinOptions()` is removed under built-in Kotlin. Use `kotlin { compilerOptions { ... } }`.
- Kotlin or javac compile errors that the collision was hiding.

So Phase 1 can state 'no collision breakages'. It should not state 'no breakages'. A share of libraries will need a second, `kotlinOptions`-style fix after the guard lands.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+| async-storage | v3.1.0 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/%40react-native-async-storage%2Basync-storage%2B3.1.0.patch |
+| nitro-image | v0.14.0 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/react-native-nitro-image%2B0.14.0.patch |
+
+<hr/>
+
+- Most libraries apply the `kotlin-android` plugin in `build.gradle`. With built-in Kotlin, this is no longer needed and can fail builds.
+- Previously, `java.srcDirs` plus `kotlin-android` covered Kotlin-only sources. With built-in Kotlin, Kotlin-only sources should be defined via `kotlin.srcDirs`.
+- `.srcDirs` is deprecated in favor of `.directories`.
+- `android.kotlinOptions` is no longer valid with built-in Kotlin; use `kotlin.compilerOptions`.
+- `fix-prefab.gradle` scripts in C++ (CMake) libraries often use removed APIs (`android.applicationVariants` / `android.libraryVariants`); migrate to `androidComponents.onVariants`.
```
gabrieldonadel: [🐧 nit] We can confirm this is real but low-frequency. In the 1002-package run, removed DSL properties (`libraryVariants`, `bootClasspath`) caused only 8 failures. Their usage sum is 0.025, max 0.013.

So this migration matters for a small tail. The Kotlin collision (445 cases) should be the first priority. Affected packages include `react-native-launch-arguments`, `react-native-worklets-core`, and `react-native-context-menu-view`.

proposals/0000-android-agp-v9-adoption.md
```diff
@@ -0,0 +1,229 @@
...
+
+The second area is AGP v9 compatibility for community libraries. Below are the common migration patterns and library-specific notes.
+
+### Community libraries
+
+Below is the matrix used during discovery:
+  
+| Library | Version | Patch link |
+| --- | --- | --- |
+| react-native-reanimated | v4.3.1 | https://github.com/hurali97/RN-AGP9-support/blob/main/patches/react-native-reanimated%2B4.4.1.patch |
```
gabrieldonadel: [🏗️ **design**] These two carry the most weight and need a different guard shape. `react-native-reanimated` (usage 0.735) and `react-native-worklets` (0.661) are the two highest-usage packages still broken by the collision.

They declare the Kotlin plugin in a `plugins {}` block. A `plugins {}` block cannot be made conditional the way `apply plugin` can. So the simple extension guard does not apply directly.

The fix is to declare with `plugins { id '...' apply false }` and then apply conditionally, or drive it through `pluginManager`. One design, verified once, unblocks more ecosystem weight than all the single-file guards combined. Worth flagging as a Phase 2 priority.

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel, line 74: The discovery pass can back this up with some data. The harness enabled both AGP 9 defaults (`android.newDsl=true`, `android.builtInKotlin=true`) and built 1002 of the most-used community packages.

Results: 364 `pass`, 281 `fail-newdsl` (fails with the flags on, builds with them off), 348 `fail-baseline`.

Full per-package results and logs are available at https://github.com/gabrieldonadel/agp-new-dsl/tree/master/logs

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel: Our data shows that this issue dominates and accounts for 96% of `fail-newdsl` causes.
 A backward-compatible guard fixes it without dropping support for older AGP. The apply must come after `apply plugin: 'com.android.library'`, so AGP has already registered its extension. This form needs no version table. It also covers AGP 10, where the opt-out is removed. 

Suggested edit to this bullet:

```suggestion
- Most libraries apply the `kotlin-android` (or `org.jetbrains.kotlin.android`) plugin unconditionally in `build.gradle`. AGP 9 registers the `kotlin` extension itself, so this now fails. The backward-compatible fix is to apply the plugin only when AGP has not already provided it: `if (project.extensions.findByName('kotlin') == null) { apply plugin: 'kotlin-android' }`, placed after `apply plugin: 'com.android.library'`.
```

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel, line 31: Much of this Phase 2 work is already in progress and could potentially be linked here.

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel, line 146: The Expo SDK 58 template already opts out of both flags in `android/gradle.properties`. This matches the Phase 1 approach in this RFC.

```suggestion
The Expo SDK 58 template opts out of both flags (`android.builtInKotlin=false`, `android.newDsl=false`) in `android/gradle.properties`, matching the Phase 1 approach above.
```

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel, line 166: [💡 suggestion] There is a faster win here that does not depend on the new library plugin. Add the extension guard to the `react-native-builder-bob` default template now. That fixes the collision (the 96% case) for every newly generated library. The guard is backward compatible, so the library-plugin migration can proceed in parallel.

Suggested edit to this bullet:

```suggestion
- This will require us to update `react-native-builder-bob` to use `com.facebook.react.library` by default. In the meantime, adding the backward-compatible Kotlin guard to the `react-native-builder-bob` template fixes the collision for every newly generated library immediately, independent of the library-plugin migration.
```

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel, line 174: Maybe we can soften this a bit. The guard removes the collision, but a second class of incompatibility appears behind it.

We rebuilt the published fixes. Of 13 fixes released to npm, only 8 build clean with the flags on. The collision (`Cannot add extension with name 'kotlin'`) is gone in all 13 logs. The 5 that still fail now fail later in the build. For example: `kotlinOptions()` is removed under built-in Kotlin (use `kotlin { compilerOptions { ... } }`), or a Kotlin/javac compile error that the collision was hiding.

```suggestion
As described in the earlier section, AGP v9 adoption is rolled out in three phases. Phase 1 is complete and should ship with React Native v0.87.x. Newly created apps do not need changes because template opt-outs are enabled by default. Existing apps upgrading to v0.87.x should apply a few manual updates described in the next section. At this stage, we expect no Kotlin plugin collision breakages with actively maintained community libraries. A separate class of incompatibility (for example the removed `android.kotlinOptions`) may still surface in some libraries once the collision is fixed, and will need its own follow-up.
```

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel, line 135: We can confirm this is real but low-frequency. In the test run, removed DSL properties (`libraryVariants`, `bootClasspath`) caused only 8 failures.

So this migration matters for a small tail. The Kotlin collision (445 cases) should be the first priority. Affected packages include `react-native-launch-arguments`, `react-native-worklets-core`, and `react-native-context-menu-view`.

proposals/0000-android-agp-v9-adoption.md
gabrieldonadel, line 114: These two carry the most weight and need a different guard shape. `react-native-reanimated` (usage 0.735) and `react-native-worklets` (0.661) are the two highest-usage packages still broken by the collision.

They declare the Kotlin plugin in a `plugins {}` block. A `plugins {}` block cannot be made conditional the way `apply plugin` can. So the simple extension guard does not apply directly.

The fix is to declare with `plugins { id '...' apply false }` and then apply conditionally, or drive it through `pluginManager`. One design, verified once, unblocks more ecosystem weight than all the single-file guards combined. Worth flagging as a Phase 2 priority.

proposals/0000-android-agp-v9-adoption.md
hurali97: Fair. But this will require a revert when adopting the `com.facebook.react.library`

proposals/0000-android-agp-v9-adoption.md
hurali97: Do we wanna add more? Like What changes Expo made to handle libraries, for eg, `kotlin-android-compat` plugin?

proposals/0000-android-agp-v9-adoption.md
hurali97: Yeah agreed. Though, they have already handled this case the same way as above.

- [Worklets](https://github.com/software-mansion/react-native-reanimated/blob/d517333c8f0e56bdd5a3a247f7ad5fb60b5e37c8/packages/react-native-worklets/android/build.gradle.kts#L14)
- [Reanimated](https://github.com/software-mansion/react-native-reanimated/blob/d517333c8f0e56bdd5a3a247f7ad5fb60b5e37c8/packages/react-native-reanimated/android/build.gradle.kts#L14)

hurali97: Do we still keep these in the RFC? I guess maybe for book keeping, like how or what were the issues recorded earlier.

proposals/0000-android-agp-v9-adoption.md
hurali97: That is correct. However, this incompatibility also happens for other cases like having defined the kotlin-only sources under `sourceSets.main.java.srcDirs`. The kotlin-only sources are now required to be added to `sourceSets.main.kotlin.directories.add`, so let's also add this one.

proposals/0000-android-agp-v9-adoption.md
hurali97: Awesome. Do you wanna write a suggestion so we can make it a part of the RFC 🙏

PR #1010, opened 2026-07-10
RFC: Media query support in StyleSheet

intergalacticspacehighway:
This PR adds an RFC to support media queries with StyleSheet and resolving them natively.

Draft PR that implements the proposed RFC - https://github.com/facebook/react-native/pull/57439

### DIFF
--- a/proposals/0000-native-media-queries.md
+++ b/proposals/0000-native-media-queries.md
@@ -74,7 +74,7 @@ Supported media features in the draft PR:
 
 ### How it works
 
-Conditional styles are authored and compiled in JavaScript, but resolved on the native side. Changing the color scheme or rotating the device re-resolves them natively in a commit hook without a React re-render.
+Conditional styles are authored and compiled in JavaScript, but resolved on the native side. Changing the color scheme or rotating the device re-resolves them natively during the shadow tree commit, without a React re-render.
 
 **1. Compilation (JS).** The style-processing path compiles each conditional value into two parts: the inline `default` (parsed and processed exactly like a normal value) plus a single `styleConditions` prop that collects the parsed conditions, keyed by property:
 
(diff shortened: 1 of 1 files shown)


proposals/0000-native-media-queries.md
```diff
@@ -0,0 +1,140 @@
...
+// authored
+{ width: { default: 120, '@media (orientation: landscape)': 300 } }
+
+// compiled
+{ width: 120, styleConditions: { width: [{ query: { orientation: 'landscape' }, value: 300 }] } }
+```
+
+- Each condition value flows through the property's normal processor (e.g. `processColor`), so once matched it behaves like any other value.
+
+**2. Resolution (native commit hook).** The `styleConditions` prop is added to the node's props. A commit hook runs on every commit and, for each node carrying conditions, evaluates its queries against the current environment and patches the matched value onto the props (or restores the unpatched props when nothing matches).
```
rubennorte: I'm not an expert on styling in RN but I think that doing this type of logic in a commit hook is a red flag. Commit hooks should mostly be used by external systems not but RN internals. I feel like this should be done directly as part of layout when committing the shadow tree (within `ShadowTree.cpp`) not from outside.

We can probably do this the same way we propagate the configuration for point scale factor, etc. from the root shadow node down to individual nodes.

cc @javache , @rozele .

intergalacticspacehighway: Thanks for reviewing. My initial instinct was to put it in `tryCommit` function, right before the layout but then i updated it after seeing `AnimationBackendCommitHook`, thinking it to be cleaner 😅. Putting it in `tryCommit` should work, will test and update the proposal/draft PR.

rubennorte: I think the shared animation backend was implemented that way because of different constraints (probably reconciling source props in React components vs. animation state for those props). We should be able to do it without them here.

intergalacticspacehighway: Tested, it works. Updated proposal - https://github.com/react-native-community/discussions-and-proposals/pull/1010/changes/9e0614b6dad42ba4b516755a9476ebcd25d59994 and PR - https://github.com/react/react-native/pull/57439/commits/6ac78fa359e25f74c97aea9d408db389da06c022.

proposals/0000-native-media-queries.md
```diff
@@ -0,0 +1,140 @@
...
+  return <View style={styles.card} />;
+}
+
+const styles = StyleSheet.create({
+  card: {
+    // A single property can carry a default plus media-query branches.
+    padding: {
+      default: 12,
+      "@media (orientation: portrait)": 24,
+      "@media (orientation: landscape)": 36,
```
satya164, line 31: I wonder if there's a more type-safe API. there's a lot of specialized syntax inside this string that's not type-checked or validated when writing code without external tools.

Maybe template literals could help here but could get messy when combining multiple.

intergalacticspacehighway: > Maybe template literals

i was thinking of something like [this](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAshAmBLAhgMQs4BXAThAzlALxQDeAUFFAPY6IQB2wmi1DAXFAORi3A7JEwLlAA+3ADbIG8fAGNkkLgG5K3MHgBmEHPgC0c6hNp75ACwgBbCF05cJiAOZnhY7vGQ4A1irVcA7ojwwGYinAAGACSkDFiWAEY6AL5gAB7h5Emq5KCQsAgo6Ji4EABqyA7wAMr8iAyOxGRqANpqVADSUHVQXhAg1Jr5SGgY2Hj4agC6EQAU0e1JnNFww0VjBM3tk0kAlOGqSc29-YMrhaMl+JPZudBnyACKWDogNXT1xGri4QAC1sNQZYFEbFPDlSpvOqOJIZKjfP7AwGke5rErgoKQ+pJKDSeBIlEXMEVDG1LH7cjkCCpXg4YBQW5DFBPF41EASCAAeXiACsIHJgAAeAAqAD5GgAlDDwNgSEACihUeAQTTILASYCcIUHMUAMjIeGQ0oYsqgmy6DEZj2eOFepMckwA-JqshTyAB6N1QIUEOkhaD4UDsqAWPAAGnpNpx9PA1HNUH+KCgAfejnIhgYAaggWCZlZ7M492ZNrznJ5fMFsQSOjFJAVVCgSpVao1UAAjAAGUNtKAAIgRAJmGmVOn0hmMOFMcgs1k4Hm8Ox7nA7mSAA). Or maybe an object API could be better? I feel syntax wise CSS-like string makes sense. I also wanted to check how StyleX handles it, seems it is much more [lenient](https://github.com/facebook/stylex/blob/e80875d25281d7d236bcda4739fae01d9f243fc8/packages/%40stylexjs/stylex/src/types/StyleXTypes.d.ts#L34) and supports any string prefixed with `@`.

rubennorte: Not necessarily advocating for this, but another option could be something like this:

```javascript
const $landscape = StyleSheet.media({
  orientation: 'landscape',
});

const $portrait = StyleSheet.media({
  orientation: 'portrait',
});

const styles = StyleSheet.create({
  padding: {
    default: 12,
    [$portrait]: 24,
    [$landscape]: 36,
  },
});
```

And possibly define some defaults:

```javascript
const styles = StyleSheet.create({
  padding: {
    default: 12,
    [StyleSheet.media.PORTRAIT]: 24,
    [StyleSheet.media.LANDSCAPE]: 36,
  },
});
```

intergalacticspacehighway: I guess we can keep the string syntax for CSS familiarity (with weak typings like StyleX) and also add `StyleSheet.media` helper. wdyt?

satya164: I guess the CSS syntax + helper makes sense. prior art in unistyles: https://v2.unistyl.es/reference/media-queries/

but also this proposal is about specifying different property values based on conditions, unlike media query where where it provides a style block. so maybe it shouldn't be based on media query syntax, but CSS if conditions, which are closer to this concept? https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/if

i was actually thinking to use a similar API in react navigation:

```js
const styles = StyleSheet.create({
  padding: {
    'media (orientation: portrait)': 24,
    'media (orientation: landscape)': 36,
    else: 12,
  },
});
```

intergalacticspacehighway: TIL `if` function. Nice, i guess long term that would be ideal. Wondering if it's still early as it [lacks support](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/if#browser_compatibility) in widely used browsers and the [proposal](https://drafts.csswg.org/css-conditional-values-1/) seems to be in draft (so it will require some polyfill for web if we support its CSS `if()` syntax). We're not doing style blocks with media query but we can aim to provide CSS like `@media` syntax for familiarity (to humans and LLMs 😅).

